PR #59, opened 2020-02-01
Better support for windows

Lej77:
I tried to use this on windows and got some compile errors so I fixed those. Then I noticed that the started tasks failed to start (they were started using `sh` instead of `cmd`). The output from `cmd` had lots of unicode replacement characters in them so I added a flag to `cmd` that made it output unicode instead and then wrote some code to parse the output as utf16 instead of utf8. Now it seems to work for simple tasks so I thought I'd create this pull request with my changes.

You will notice that the `daemon/task_handler.rs` file just returns errors when signals are used. So something better should probably be done there but otherwise it should hopefully work. Also I'm not 100% sure about the config paths but they seem to be okay.

### DIFF
--- a/shared/log.rs
+++ b/shared/log.rs
@@ -1,5 +1,7 @@
 use ::anyhow::Result;
+use ::byteorder::{LittleEndian, ReadBytesExt};
 use ::log::error;
+use ::std::borrow::Cow;
 use ::std::fs::{remove_file, File};
 use ::std::io::prelude::*;
 use ::std::path::{Path, PathBuf};
@@ -32,6 +34,30 @@ pub fn get_log_file_handles(task_id: usize, settings: &Settings) -> Result<(File
     Ok((stdout, stderr))
 }
 
+fn command_data_to_text(mut data: &[u8]) -> Cow<str> {
+    if cfg!(windows) {
+        // On windows we run the command using "cmd" with a flag that makes it output Unicode (UTF16).
+
+        let is_odd = data.len() % 2 == 1;
+        let mut buffer = Vec::with_capacity(data.len() / 2 + if is_odd { 1 } else { 0 });
+
+        type CmdUtf16Endian = LittleEndian;
+
+        while data.len() > 1 {
+            buffer.push(data.read_u16::<CmdUtf16Endian>().unwrap());
+        }
+        if is_odd {
+            let extra = [*data.last().unwrap(), 0];
+            let mut extra = &extra as &[u8];
+            buffer.push(extra.read_u16::<CmdUtf16Endian>().unwrap());
+        }
+
+        Cow::from(String::from_utf16_lossy(&buffer))
+    } else {
+        String::from_utf8_lossy(data)
+    }
+}
+
 /// Return the content of temporary stdout and stderr files for a task
 pub fn read_log_files(task_id: usize, settings: &Settings) -> Result<(String, String)> {
     let (mut stdout_handle, mut stderr_handle) = get_log_file_handles(task_id, settings)?;
@@ -41,8 +67,8 @@ pub fn read_log_files(task_id: usize, settings: &Settings) -> Result<(String, St
     stdout_handle.read_to_end(&mut stdout_buffer)?;
     stderr_handle.read_to_end(&mut stderr_buffer)?;
 
-    let stdout = String::from_utf8_lossy(&stdout_buffer);
-    let stderr = String::from_utf8_lossy(&stderr_buffer);
+    let stdout = command_data_to_text(&stdout_buffer);
+    let stderr = command_data_to_text(&stderr_buffer);
 
     Ok((stdout.to_string(), stderr.to_string()))
 }
--- a/shared/log.rs
+++ b/shared/log.rs
@@ -1,7 +1,6 @@
 use ::anyhow::Result;
 use ::byteorder::{LittleEndian, ReadBytesExt};
 use ::log::error;
-use ::std::borrow::Cow;
 use ::std::fs::{remove_file, File};
 use ::std::io::prelude::*;
 use ::std::path::{Path, PathBuf};
@@ -34,7 +33,8 @@ pub fn get_log_file_handles(task_id: usize, settings: &Settings) -> Result<(File
     Ok((stdout, stderr))
 }
 
-fn command_data_to_text(mut data: &[u8]) -> Cow<str> {
+/// Convert stdout or stderr data from a spawned task to a string.
+pub fn process_output_to_text(mut data: &[u8]) -> String {
     if cfg!(windows) {
         // On windows we run the command using "cmd" with a flag that makes it output Unicode (UTF16).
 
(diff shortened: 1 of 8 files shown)


Nukesor:
Wow! Thanks a lot!

I'll take a proper look at it as soon as I get some spare time!

One thing you can already do:
Add this back to the `.github/workflows/{test.yml,package-binary.yml}` matrix section:

```
matrix:
...
    os: [ubuntu-latest, windows-latest, macos-latest]
    include:
            - os: windows-latest
              client_name: pueue.exe
              client_release_name: pueue-windows-amd64.exe
              daemon_name: pueued.exe
              daemon_release_name: pueued-windows-amd64.exe
```

Nukesor:
I just saw, that the whole `include` section is not longer necessary in `test.yml`.
If you want you can remove the `include:` section from `test.yml` while you're at it.

shared/log.rs
```diff
@@ -32,6 +34,30 @@ pub fn get_log_file_handles(task_id: usize, settings: &Settings) -> Result<(File
...
+        while data.len() > 1 {
+            buffer.push(data.read_u16::<CmdUtf16Endian>().unwrap());
+        }
+        if is_odd {
+            let extra = [*data.last().unwrap(), 0];
+            let mut extra = &extra as &[u8];
+            buffer.push(extra.read_u16::<CmdUtf16Endian>().unwrap());
+        }
+
+        Cow::from(String::from_utf16_lossy(&buffer))
```
Nukesor: Since the Cow is reverted to a String anyway after being returned, just return a String directly instead of `Cow<str>`.
This removes a little bit of unnecessary complexity.

shared/log.rs
```diff
@@ -32,6 +34,30 @@ pub fn get_log_file_handles(task_id: usize, settings: &Settings) -> Result<(File
     Ok((stdout, stderr))
 }
 
+fn command_data_to_text(mut data: &[u8]) -> Cow<str> {
```
Nukesor: I'm not 100% sure about the name `command_data_to_text`, `command_data` seems a little vague.
Maybe something like `process_output_to_text`?

Lej77:
I agree that the `command_data_to_text` functions name wasn't that great and I change it to the one you suggested instead. The switch from `Cow<str>` to `String` definitely simplifies the code a bit as well. For the streaming I used use the `process_output_to_text` function to convert the data to text, from what I understand that should be all that was needed.

## `powershell` current directory path issues

Earlier I also added a [`Escape task command correctly on windows using powershell`](https://github.com/Nukesor/pueue/pull/59/commits/1ddececf9e7871e5b7e75f498eca9e9d2a7bc4e3) commit since `cmd` has some really weird argument escaping. The current code should work but there is other ways that the issue could have been solved. The issue is that `cmd` expects all quotes to not be escaped (don't convert `"` to `\"`) but Rust `std::process::Command` API doesn't provide anyway to skip that conversion. So to fix that I instead pass the task argument to `powershell` instead since that expects the standard string escaping. Then I use `powershell` to start `cmd` since its API doesn't auto escape arguments. 

One alternative could be to just use `powershell` for tasks instead of `cmd`, I choose not to do that since I think more people are used to `cmd` than `powershell` but its an option. Another alternative would be to start `cmd` some other way, for example via a temporary `VB Script` file, via a crate that provides a raw process start API or by writing the `cmd` command to a temporary batch file that is then started. I think the current solution is good enough but there is certainly alternatives.

Using `powershell` to start `cmd` does have some disadvantages:
1. It might have a performance overhead (though this should hopefully be negligible)
2. From what I have observed `powershell` has issues with current directory paths that contain `[` or `]` characters (this has caused me issues in the past). When it is started with a current directory that contains those characters its current directory will instead default to `"C:\WINDOWS\System32\WindowsPowerShell\v1.0"`. 

We could work around the second issue by setting the current directory with a `cmd` command like `cmd /U /S /C start /D "C:\Users\Henrik\Downloads\New folder [hello] (5)" /Wait /B cmd /U /S /C "INSERT_REAL_TASK"` but the task command would need to be escaped in some way and it doesn't some things probably couldn't be escaped correctly. Another workaround would be to use the `powershell` start command when starting `cmd` and specifying the current directory. That seems to work but the command will error out if the current directory path doesn't exist. Here is some code for generating such a `powershell` command:
```rust
.arg(format!(
    "start -NoNewWindow -Wait -WorkingDirectory '{}' cmd /U,/S,/C,'{}'",
    task.path.replace('\'', "''").replace('[', "`[").replace(']', "`]"),
    task.command.replace('\'', "''")
));
```

Anyway I don't think its worth bothering about, if `powershell` can ignore it then we probably can as well.

Nukesor:
Ok. This seems to be way more complicated than I expected.
First of all:
Is the syntax used in Powershell and cmd very different? If that's not the case, just using Powershell would probably be the easiest mode of operation.

Regarding the disadvantages:
1. Performance overhead should really be negligible
2. I guess there exists some kind of library for proper escaping for Powershell somewhere. If we want to be safe, we should include this. Anyway, I think this is something for a different PR. Let's get the basics up and running for now.
3. This could be annoying, but hopefully nothing like this should happen. If we decide to not use `cmd` this doesn't even become a problem.

I guess just using PowerShell is probably the easiest solution for all problems?

Lej77:
I think it works well enough currently. Still your right that just using `powershell` could be easier, I just started with `cmd` since that seemed the go to solution. 

From what I read about `powershell` strings there should hopefully not be any characters that are unescaped but `powershell` seems to be inconsistent with what characters need to be escaped depending on where the string is going to be used. You can see that in the example code in my previous comment I replaced more characters when the string was going to be used as a path. Those replacements should not be done for the command string.

Since the `powershell` command should never fail the third issue shouldn't be a problem. I could change the code to use `powershell` directly but then I would need to figure out how to get it to output a sensible encoding (does it do this by default?) so that would be some extra work. Still it could simplify the code slightly since the utf-16 string decoding would probably not be needed. I think it is pretty good as it is but if you want to I can change it to only use `powershell`. Maybe there could be a config for which shell should be used? But that is probably not something we should do in this PR, we could leave that for future improvements.

Lej77:
I noticed that the `cmd` encoding didn't affect external programs so running a task like `cargo --help` would produce gibberish. For that reason and the ones you discussed above I switched to using `powershell` instead of `cmd` for commands.

It would be nice if we could detect if `pwsh` is installed and use that instead when it is to fix those pesky current directory issues but otherwise I think that this solves all the issues I brought up.

Nukesor:
Well, let's merge this first. We will see if any problems arise.
I'll add a notice to Readme, that Windows support is still very experimental :)

Lej77:
Sounds good! There isn't really that much windows specific code (this PR is overall quite small) so I think everything should mostly work. I'm quite happy with the `powershell` solution and it shouldn't really be an issue for most users. So you probably don't need to specify that it is very experimental but you could mention that it is less tested and there might be some bugs.

PR #64, opened 2020-02-03
Allow compilation on FreeBSD

twillingham:
Currently, the app has functions for Linux, MacOS, and Windows configuration directories. The linux directories also work on Freebsd, and Pueue runs fine on FreeBSD 12.1 once this change is added.

I'm not familiar enough with Rust yet to know if `target_family="unix"` might be a better approach to this, since I am not sure if that is picked up by MacOS which has requirements for different paths.

### DIFF
--- a/shared/settings.rs
+++ b/shared/settings.rs
@@ -107,12 +107,12 @@ fn gen_random_secret() -> String {
     secret
 }
 
-#[cfg(target_os = "linux")]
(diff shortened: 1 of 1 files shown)


Nukesor:
LGTM

`unix` as target_family would probably break the macOS build. Let's keep it like this and specify paths for each target system explicitly. Better safe than sorry.

Nukesor:
Thanks for your contribution!

PR #74, opened 2020-02-26
Format code using 'cargo fmt'

Atul9:


Atul9:
Hi @Nukesor can you please share why was this PR closed?

Nukesor:
I usually cargo fmt whenever I work on the project. Looks like I forgot to do so in the last 1-2 commits.

However, I'm not a huge fan of automated formatting commits by bots. If you want to do a proper contribution, add some kind of hook to the repository instead of harvesting automated open source contributions that'll likely reoccur at some point of time.

Solve the problem for good instead of offering automated quickfixes.

Issue #72, opened 2020-02-22
Delay task enqueue


taylor1791:
I often need to schedule certain tasks. e.g. `scp` this file tomorrow. Having this feature built in would be really nice.

```
pueue add -- scp file user@host:/srv/www/releases/
pueue stash <task-id>
pueue enqueue --delay-until '2020-04-01T00:00:00' <task-id>
```

If combined with the tags feature mentioned in !44, one could have a finer control over when a task was actually completed. Imagine creating a tag for short lived, high urgency tasks. 

I have been working around it currently, by using `sleep 3600 && pueue add ...` in a tmux session, but having 7 of those (one for every day of the week is really annoying). I have tried using `at`, but I can't do that because it requires disabling SIP on MacOS these days. Ug.

By adding this feature it might also require being able to add a stashed task. I could imagine this being `pueue add --stash -- ...`. This would avoid executing it immediately if the queue happened to be empty.

To create a really nice user experience, an option like `--delay-until now + 2 days` or `--delay-until tomorrow` would be really nice. This syntax was taken from the `at` command. That being said, I would be more than happy with any format.

I looked into the source and I am willing and feel pretty confident I could make the contribution after agreeing on a design.

Nukesor:
I really like the idea. It should have a quite low impact on the code base as well.

Until the feature is added, you can simply do `pueue add "sleep 3600 && your_command"`. This should remove the necessity for tmux.

1. A `--stash` flag would even be convenient without the delay feature.
2. Adding `--delay-until` it to `add` and `enqueue` sounds good :). I agree that having tags would be super convenient to ensure that the task will be started at this point of time. Alternatively one could also add a `--start-at` flag, which forces the start of the task at a specific point of time. Although, this might be something for another PR.
2.1. However, I don't know yet how this should be shown in the UI. Another column which is only displayed, if there is at leas one delayed task?  
3. For parsing the input, [humantime](https://github.com/tailhook/humantime) looks like a good solution. I really don't want to add the overhead of maintaining a custom written parser to Pueue.

Feel free to start hacking :)! I'll try to answer all questions you might have

taylor1791:
Here is the plan, broken into decisions. Let me know what you think. My potential concerns are listen at the end.

1. **`pueue add`**. Add `--delay <str>` to `add`. It will require the `--stash` option.
1. **`pueue enqueue`**. Add `--delay <str>` argument.
[...]

PR #75, opened 2020-03-07
Support delayed tasks

taylor1791:
Issue #72 laid out a design to support delayed task enqueuing. This PR implements that design.

Rust is not my native tongue and I learned more about rust by reading the code. More than likely there were things I overlooked, so feel free to request any and all changes that cross your mind.


daemon/task_handler.rs
```diff
@@ -158,6 +159,30 @@ impl TaskHandler {
         state.change_status(task_id, TaskStatus::Running);
     }
 
+    /// As time passes, some delayed tasks may need to be enqueued.
+    /// Gather all stashed tasks and enqueue them if it is after the task's enqueue_at
+    fn check_stashed(&mut self) {
+        let stashed_tasks = {
+            self.state.lock().unwrap().tasks_in_statuses(None, vec![TaskStatus::Stashed]).0
```
Nukesor: This function became a little more complicated than it would have needed to.
You can get the mutex on the state once and keep it until everything is finished. This is also important to ensure, that the state won't be manipulated in another thread and thereby gives us a few nice guarantees.

I just did a quick rewrite for you to understand how it may look nicer :)
```
    /// As time passes, some delayed tasks may need to be enqueued.
    /// Gather all stashed tasks and enqueue them if it is after the task's enqueue_at
    fn check_stashed(&mut self) {
        let mut state = self.state.lock().unwrap();
        let stashed_tasks = state.tasks_in_statuses(None, vec![TaskStatus::Stashed]).0;

        for task_id in stashed_tasks {
            // We can unwrap, since we got a mutex on the state and know, that the ids are valid.
            let task = state.tasks.get_mut(&task_id).unwrap();

            if let Some(time) = task.enqueue_at {
                if time <= Local::now() {
                    info!("Delayed task enqueued: {:?}", task);

                    task.status = TaskStatus::Queued;
                    task.enqueue_at = None;
                }
            }
        }
    }
```

shared/task.rs
```diff
@@ -55,6 +62,7 @@ impl Task {
             id: 0,
             command: task.command.clone(),
             path: task.path.clone(),
+            enqueue_at: task.enqueue_at,
```
Nukesor: The `from_task` function is used when restarting an existing task.
Copying task.enqueue_at doesn't make much sense in here. Just use `None`

daemon/instructions.rs
```diff
@@ -131,14 +136,23 @@ fn enqueue(message: EnqueueMessage, state: &SharedState) -> Message {
...
+                state.change_status(*task_id, TaskStatus::Queued);
+            }
         }
 
         (matching, mismatching)
     };
 
-    let message = "Tasks are enqueued";
+    let message = if message.enqueue_at.is_some() {
+        "Tasks enqueueing delayed"
```
Nukesor: I think it would be super awesome to get something like:
`Tasks will be enqueued at "Here comes the datetime"`
A good old visual check without calling `pueue status` could be handy.

If you use formatting more often, something like a getter for formatted dates of enqueue, which returns something like `Option<String>` , could be useful. But maybe that's overkill. Do as you wish :)  

Plus `enqueueing` is hell of a word :laughing:

daemon/instructions.rs
```diff
@@ -36,13 +36,18 @@ pub fn handle_message(message: Message, sender: &Sender<Message>, state: &Shared
...
         TaskStatus::Queued
     };
 
-    let task = Task::new(message.command, message.path, starting_status);
+    let task = Task::new(
+        message.command,
+        message.path,
+        starting_status,
+        message.enqueue_at
+    );
```
Nukesor: I cannot comment on the correct line, but it would be awesome to get a custom success message for delayed tasks when adding as well :)

daemon/instructions.rs
```diff
@@ -131,14 +136,23 @@ fn enqueue(message: EnqueueMessage, state: &SharedState) -> Message {
...
         }
 
         (matching, mismatching)
     };
 
-    let message = "Tasks are enqueued";
+    let message = if message.enqueue_at.is_some() {
+        "Tasks enqueueing delayed"
+    } else {
+        "Tasks are enqueued"
```
Nukesor: Ok. Here's what I'm thinking.

If you add an `enqueue_at` to a task and then `enqueue` or `start` it manually afterwards, before it has been enqueued automatically, the `enqueue_at` field will still be a `Some<DateTime>` instead of a `None`.
I guess this wouldn't be bad, since you only check `Stashed` tasks in `check_stashed`. However, it would make me feel better, if you could set `task.enqueue_at = None` right here.

While reviewing my own code I also found that the passthrough of a cloned task into `start_process` is super ugly. I guess I'll refactor this as soon as your PR is merged. I'll then add the `task.enqueue_at = None` to `start_process` myself :)

daemon/task_handler.rs
taylor1791: That is much better, thanks!

shared/task.rs
taylor1791: Updated.

daemon/instructions.rs
taylor1791: Added.

daemon/instructions.rs
taylor1791: Done

daemon/instructions.rs
taylor1791: I fixed the `enqueue` command by always setting it. It doesn't solve the same problem with the `start` command. However, I think you were going to address that in your refactor? Am I correct?

taylor1791:
I think I did everything you asked. Let me know if I misinterpreted anything.

Regarding being added to `CONTRIBUTORS.md`. I'll let you decide if you think my contributions warrant creating and adding me to it. I currently have no preference.

daemon/instructions.rs
Nukesor: Jep that's correct :)

Nukesor:
Oook. LGTM. 
I'll push a new release as soon as I find some time to do a little bit of refactoring on my old code :)

Thanks again for contributing!

PR #81, opened 2020-04-10
Add a dependencies system

tinou98:
It adds the possibility to set dependencies between tasks, it adds a way to control parallelism

For instance :
```
$ pueue add -- git pull
New task added (id 1).

$ pueue add --after 1 -- make -j2
```
This will ensure that `make` is run after `git pull` has succeeded.
You can also add multiple dependencies, to create complex pipeline.

### DIFF
--- a/client/cli.rs
+++ b/client/cli.rs
@@ -28,7 +28,7 @@ pub enum SubCommand {
         delay_until: Option<DateTime<Local>>,
 
         /// Start the task once all specified tasks have successfully finished.
-        /// If one of the dependencies failed, this task will be marked as failed
+        /// As soon as one of the dependencies fails, this task will be fail as well
         #[structopt(name = "after", long)]
         dependencies: Vec<usize>,
     },
--- a/client/cli.rs
+++ b/client/cli.rs
@@ -26,6 +26,10 @@ pub enum SubCommand {
         /// Delays enqueueing the task until <delay> elapses. See enqueue for accepted formats.
         #[structopt(name = "delay", short, long, conflicts_with = "immediate", parse(try_from_str=parse_delay_until))]
         delay_until: Option<DateTime<Local>>,
+
+        /// Only start the task after <after> task has finished
+        #[structopt(name = "after", long)]
+        dependencies: Vec<usize>,
     },
     /// Remove tasks from the list.
     /// Running or paused tasks need to be killed first.
--- a/client/cli.rs
+++ b/client/cli.rs
@@ -27,7 +27,8 @@ pub enum SubCommand {
         #[structopt(name = "delay", short, long, conflicts_with = "immediate", parse(try_from_str=parse_delay_until))]
         delay_until: Option<DateTime<Local>>,
 
-        /// Only start the task after <after> task has finished
+        /// Start the task once all specified tasks have successfully finished.
+        /// If one of the dependencies failed, this task will be marked as failed
         #[structopt(name = "after", long)]
         dependencies: Vec<usize>,
     },
--- a/shared/state.rs
+++ b/shared/state.rs
@@ -43,6 +43,9 @@ impl State {
     }
 
     pub fn get_next_task_id(&mut self) -> Option<usize> {
+        // A runnable task:
+        // - is in Queued state
+        // - has all task in Done state
         return self
             .tasks
             .iter()
@@ -300,5 +303,32 @@ impl State {
         }
 
         Ok(())
+    }
+
+    /// Checkek that any Queued tasks dont have any failed dependencies, otherwise marked it as failed
+    pub fn update_dependencies(&mut self) {
+        let has_failed_deps: Vec<_> = self
+            .tasks
+            .iter()
+            .filter(|(_, task)| task.status == TaskStatus::Queued)
+            .filter_map(|(id, task)| {
+                let failed = task
+                    .dependencies
+                    .iter()
+                    .flat_map(|id| self.tasks.get(id))
+                    .filter(|task| task.is_errored())
+                    .map(|task| task.id)
+                    .next();
+
+                failed.map(|f| (*id, f))
+            })
+            .collect();
+
+        for (id, failed) in has_failed_deps {
+            if let Some(task) = self.tasks.get_mut(&id) {
+                task.status = TaskStatus::Failed;
+                task.stderr = Some(format!("Dependent task {:?} has failed", failed));
+            }
+        }
     }
 }
--- a/shared/state.rs
+++ b/shared/state.rs
@@ -42,10 +42,11 @@ impl State {
         self.max_id - 1
     }
 
+    /// Search and return the next runnable task.
+    /// A runnable task:
+    /// - is in Queued state
+    /// - has all its dependencies in `Done` state
     pub fn get_next_task_id(&mut self) -> Option<usize> {
-        // A runnable task:
-        // - is in Queued state
-        // - has all task in Done state
         return self
             .tasks
             .iter()
@@ -305,8 +306,8 @@ impl State {
         Ok(())
     }
 
-    /// Checkek that any Queued tasks dont have any failed dependencies, otherwise marked it as failed
-    pub fn update_dependencies(&mut self) {
+    /// Ensure that no `Queued` tasks dont have any failed dependencies, otherwise set their status to `Failed`.
+    pub fn check_failed_dependencies(&mut self) {
         let has_failed_deps: Vec<_> = self
             .tasks
             .iter()
--- a/shared/state.rs
+++ b/shared/state.rs
@@ -43,12 +43,18 @@ impl State {
     }
 
     pub fn get_next_task_id(&mut self) -> Option<usize> {
-        for (id, task) in self.tasks.iter() {
-            if task.status == TaskStatus::Queued {
-                return Some(*id);
-            }
-        }
-        None
+        return self
+            .tasks
+            .iter()
+            .filter(|(_, task)| task.status == TaskStatus::Queued)
+            .filter(|(_, task)| {
+                task.dependencies
+                    .iter()
+                    .flat_map(|id| self.tasks.get(id))
+                    .all(|task| task.status == TaskStatus::Done)
+            })
+            .next()
+            .map(|(id, _)| *id);
     }
 
     pub fn change_status(&mut self, id: usize, new_status: TaskStatus) {
--- a/client/message.rs
+++ b/client/message.rs
@@ -14,6 +14,7 @@ pub fn get_message_from_opt(opt: &Opt) -> Result<Message> {
             start_immediately,
             stashed,
             delay_until,
+            dependencies
         } => {
             let cwd_pathbuf = current_dir()?;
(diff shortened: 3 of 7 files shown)


shared/state.rs
```diff
@@ -43,12 +43,18 @@ impl State {
     }
```
Nukesor, line 44: Please add a few lines of comments. This function just became a lot more complex and deserves a little explanation :)

shared/state.rs
```diff
@@ -43,12 +43,18 @@ impl State {
...
-        None
+        return self
+            .tasks
+            .iter()
+            .filter(|(_, task)| task.status == TaskStatus::Queued)
+            .filter(|(_, task)| {
+                task.dependencies
+                    .iter()
+                    .flat_map(|id| self.tasks.get(id))
+                    .all(|task| task.status == TaskStatus::Done)
```
Nukesor, line 54: I feel like we need to handle the case that tasks fail. (`Killed`, `Failed`).
Right now this would result in tasks being stuck in `Queued` forever without obvious reason.

1. Either show the tasks the affected tasks is depending on in the `pueue status` overview. (of course this column should only be displayed, if there is any active dependency in the current state)
2. Automatically fail tasks that depend on other failed tasks. This should also result in some kind of error message output that can be checked via `log`.

client/cli.rs
```diff
@@ -26,6 +26,10 @@ pub enum SubCommand {
         /// Delays enqueueing the task until <delay> elapses. See enqueue for accepted formats.
         #[structopt(name = "delay", short, long, conflicts_with = "immediate", parse(try_from_str=parse_delay_until))]
         delay_until: Option<DateTime<Local>>,
+
+        /// Only start the task after <after> task has finished
```
Nukesor, line 30: `after <after>` is somewhat confusing.
Maybe
`/// Only start the task after all specified tasks have finished.`

+ If we decide to auto fail the task if a parent task fails, this should be mentioned in this doc string as well.

shared/state.rs
Nukesor: After thinking a little bit about this, automatic failure of all `children` seems to be the best way.

As a consequence, we should also think about a `--dependants/-d` flag to `restart`. If this flag is set, all tasks that depend on the task to restart, will be restarted as well.

Nukesor:
```
        5
       / \
      3   4
     / \
    1   2
```

Example:
Task 1 and 4 fail.
Task 2 succeeds.
Task 3 and 5 fail automatically.

Now do a `pueue restart 1 -d`:
```
        9
       / \
      7   8?
     /
    6
```


- Restart task 1 with id 6
- Restart task 3 as task 7, depending on task 6
- Restart task 6 as task 9, depending on task 7
- Should task 4 be restarted as well? It's part of the dependency graph, but it's also a leaf (and thereby a starting point) of the graph.

tinou98: It seems to be interesting, but we may have some weird case like :
```
  4
 / \
2   3
 \ /
  1
```
In this case task 1 need to be restarted only once.

I will have a look at this feature, maybe we should create another pull request / issue

client/cli.rs
tinou98: Updated to : 
`Start the task once all specified tasks have successfully finished. If one of the dependencies failed, this task will be marked as failed`

shared/state.rs
```diff
@@ -43,6 +43,9 @@ impl State {
     }
 
     pub fn get_next_task_id(&mut self) -> Option<usize> {
+        // A runnable task:
```
Nukesor, line 46: Please do this as a function docstring instead of inline comments.
thanks :)

shared/state.rs
```diff
@@ -301,4 +304,31 @@ impl State {
 
         Ok(())
     }
+
+    /// Checkek that any Queued tasks dont have any failed dependencies, otherwise marked it as failed
+    pub fn update_dependencies(&mut self) {
```
Nukesor, line 309: `update_dependencies` is probably a bad name for this function. Something like `check_{for?}_failed_dependencies` is probably better.

shared/state.rs
```diff
@@ -301,4 +304,31 @@ impl State {
 
         Ok(())
     }
+
+    /// Checkek that any Queued tasks dont have any failed dependencies, otherwise marked it as failed
```
Nukesor, line 308:
```suggestion
    /// Ensure that no `Queued` tasks have any failed dependencies, otherwise set their status to as `Failed`.
```

shared/state.rs
```diff
@@ -301,4 +304,31 @@ impl State {
...
+    pub fn update_dependencies(&mut self) {
+        let has_failed_deps: Vec<_> = self
+            .tasks
+            .iter()
+            .filter(|(_, task)| task.status == TaskStatus::Queued)
+            .filter_map(|(id, task)| {
```
Nukesor, line 314: As far as I can see, you only look for single level dependencies.

Scenario:
```
a -> b -> c
```
`a` fails, `b` detects `a` as dependency and fails.
`c` however didn't detect `b` as a failed dependency before.
`c` is started.

As I said, we will probably need proper dependency graph handling or you have to implement a simple recursive lookup for now.

client/cli.rs
```diff
@@ -27,7 +27,8 @@ pub enum SubCommand {
         #[structopt(name = "delay", short, long, conflicts_with = "immediate", parse(try_from_str=parse_delay_until))]
         delay_until: Option<DateTime<Local>>,
 
-        /// Only start the task after <after> task has finished
+        /// Start the task once all specified tasks have successfully finished.
+        /// If one of the dependencies failed, this task will be marked as failed
```
Nukesor, line 31:
```suggestion
        /// As soon as one of the dependencies fails, this task will be fail as well
```

shared/state.rs
tinou98: This will work : 
`a` fails, `b` detects `a` as dependency and fails.
`c` however didn't detect `b` as a failed dependency before.
`c` is **not** started, because `b` is not in `Done` state.
`c` detects that `b` failed and is now failed as well.

It takes 1 more internal tick, but it won't start task by mistake.

Nukesor: Yess, you're right. Sorry, forgot that tasks won't start until their dependencies are finished

Nukesor:
Ok LGTM :)
I'll make a few test runs myself. If everything checks out, this can be merged.

Nukesor:
Ok.
After playing around with the new feature, one thing that's definitely missing is information about dependencies in `pueue status`. Otherwise there is no indication whether you managed to setup your dependencies correctly.

The new column should be only displayed when there are tasks that have dependencies.

tinou98:
I totally agree with that, and for me it would be really be beneficial to refactor the Status type, so we can express more detail, something like : 
```rust
pub enum WaitingStatus {
    /// Waiting for a free slot, limited by parallelism
    Queued,

    /// Waiting to be started at `enqueue_at`
    Stashed,

    /// Blocked by dependencies
    WaitingForDependencies(Vec<usize>),

    /// Used while the command of a task is edited (to prevent starting the task)
    Locked,
}

pub enum RunningStatus {
    Running,
    Paused,
}

pub enum EndStatus {
    Done,
    Failed {
        exit_code: i32,
    },
    Killed,
}

pub enum TaskStatus {
    Waiting(WaitingStatus),
    Running {
        status: RunningStatus,
        start: DateTime<Local>,
    },
    Ended {
        status: EndStatus,
        start: DateTime<Local>,
        end: DateTime<Local>,
        stdout: String,
        stderr: String,
    },
}
```

Nukesor:
Hmm. This is a pretty big change, but I see how this can be beneficial.

If you don't mind, I would like to code this change myself, so I get a feeling how this changes the flow and code-style of pueue. After all, this is a pretty substantial refactoring.

Nukesor:
Ok. After a brief investigation, I would say, that this change is unrealistic.

Changing the complete layout of `Task` would result in a probably 60% rewrite of the whole project.
On top of that, I don't see enough benefits from this.
I agree that the data structure might be more sane, but ergonomics will suffer from those changes.

Instead of one lookup/change, we'll have to do nested lookups and nested access (match => match => ...) for changing stuff pretty much everywhere. I suspect that the code will become significantly bloated by this. 

If you would still like to give it a try, feel free to implement those changes. However, don't be surprised, if the PR won't be accepted.
However, I'll still give it a proper consideration. It still possible that I'm completely wrong about this.

This is the updated Task file I came up with:
```

#[derive(Clone, Display, Debug, Serialize, Deserialize, PartialEq)]
pub enum Waiting {
    /// Waiting for a execution by the TaskManager
    Queued,

    /// Waiting to be enqueued manually or
    /// automatically via `enqueue_at` by the TaskManager
    Stashed,

    /// Blocked by dependencies
    /// As soon as all dependencies are done, the task can be started
    WaitingForDependencies(Vec<usize>),

    /// Used while a task is being edited (to prevent the task being started)
    Locked,
}

#[derive(Clone, Display, Debug, Serialize, Deserialize, PartialEq)]
pub enum Running {
    /// The task has been started and is currently running
    Running,
    /// The task has been started, but manually paused afterwards
    /// Won't be started unless the user manually does so.
    Paused,
}

#[derive(Clone, Display, Debug, Serialize, Deserialize, PartialEq)]
pub enum Finished {
    /// The task finished successfully
    Done,
    /// The task somehow failed
    Failed,
    /// The task has been actively killed
    Killed,
}

#[derive(Clone, Display, Debug, Serialize, Deserialize, PartialEq)]
pub enum TaskState {
    Waiting {
        status: Waiting,
        enqueue_at: Option<DateTime<Local>>,
    },
    Running {
        status: Running ,
        start: DateTime<Local>,
    },
    Finished {
        status: Finished,
        exit_code: i32,
        start: DateTime<Local>,
        end: DateTime<Local>,
        stdout: String,
        stderr: String,
    },
}

/// Representation of a task.
/// start will be set the second the task starts processing.
/// exit_code, output and end won't be initialized, until the task has finished.
/// The output of the task is written into seperate files.
/// Upon task completion, the output is read from the files and put into the struct.
#[derive(Clone, Debug, Serialize, Deserialize)]
pub struct Task {
    pub id: usize,
    pub command: String,
    pub path: String,
    pub state: TaskState,
    pub prev_state: TaskState,
}

impl Task {
    pub fn new(
        command: String,
        path: String,
        starting_state: TaskState,
    ) -> Task {
        Task {
            id: 0,
            command: command,
            path: path,
            state: starting_state.clone(),
            prev_state: starting_state,
        }
    }

    pub fn from_task(task: &Task) -> Task {
        Task {
            id: 0,
            command: task.command.clone(),
            path: task.path.clone(),
            state: TaskState::Waiting{ status: Waiting::Queued, enqueue_at: None},
            prev_state: TaskState::Waiting{ status: Waiting::Queued, enqueue_at: None},
        }
    }

    //pub fn is_running(&self) -> bool {
    //    return self.state == TaskState::Running{_};
    //}

    //pub fn is_done(&self) -> bool {
    //    return self.state == TaskState::Finished;
    //}

    //pub fn is_waiting(&self) -> bool {
    //    return self.state == TaskState::Waiting;
    //}
}
```

Issue #82, opened 2020-04-14
[BUG] excessive memory and CPU usage with large outputs


sourcefrog:
I'm using pueue to queue up some large file copies -- and it's pretty nice!

However I notice after a while, it's taking several seconds of pueued CPU time to `pueue add date` a new task. I would have expected it's just adding it to some kind of in-memory or on-disk queue, and the time should be proportional to the fairly short command line.

I also see `pueued` seems to be using a lot of memory for what it's doing:

```
  PID TTY      STAT   TIME  MAJFL   TRS   DRS   RSS %MEM COMMAND 
16885 pts/11   SNl+   2:54     71  1930 1043077 779556 13.3 pueued
```

I don't know how to reproduce it, but maybe it's connected to some of these commands producing a lot of output? Ideally it would just stay on files on disk and not be in memory.

This is pueue 0.3.1 from Cargo, on Debian Buster.

sourcefrog:
Yep, it looks like (in https://github.com/Nukesor/pueue/blob/f85e5b4b6e3e15c911c7087fd2386aee2190f246/shared/task.rs#L24-L36) the `Task` struct includes a String of all the stdout/stderr from the task, and also the struct is `Clone`. 

I think that means when inserting entries into the `state.tasks` BTreeMap (https://github.com/Nukesor/pueue/blob/f85e5b4b6e3e15c911c7087fd2386aee2190f246/shared/state.rs#L17-L23) it'll end up not only holding the output in memory, but possibly copying it many times. 

I guess I've learned I should redirect bulk output to a file, not let pueue hold it. If that's just the intention perhaps it should be in the instructions. 

However perhaps it's not uncommon to want to run something that does produce a lot of output within a queue manager. So could I suggest perhaps 

 - keep output on disk not in memory?
 - or, at least, store them in `Cow<String>` rather than copying every time the `Task` is copied?

Nukesor:
Yep. That's a problem that already occurred in another form.

One person did pipe the output of `yes` to stdout instead into the program and completely filled their harddrive/ram with `y` in the log output :D.

To be honest, I don't think that this is a problem that should be handled by pueue. Maybe we can add some in-memory compression. In general, I think it's way too convenient to have your logs ready to use. If we want to add another step and write to/read from disk, we might have a considerable delay when issuing commands.

Plus, in my opinion, piping gigabytes of data to stdout can be considered as an edgecase.

[...]

PR #84, opened 2020-04-18
Avoid repeatedly cloning large output strings

sourcefrog:
Hold stdout and stderr in Arc refcounts so that the task can be cloned
or sent between threads without cloning the potentially-large
buffers.

In particular, because the logs were previously held inline in the
`State`, this makes manipulation of the BTree faster because the nodes
are (much) smaller.

Helps with https://github.com/Nukesor/pueue/issues/82.


Nukesor:
Closed in favor to #88

PR #86, opened 2020-04-19
[WIP] Refactor state

tinou98:
First draft of the refactor of `State` enum (as discussed in https://github.com/Nukesor/pueue/pull/81#issuecomment-615113064)

### DIFF
--- a/shared/task.rs
+++ b/shared/task.rs
@@ -33,6 +33,7 @@ pub struct Task {
     pub stderr: Option<String>,
     pub start: Option<DateTime<Local>>,
     pub end: Option<DateTime<Local>>,
+    pub dependencies: Vec<usize>,
 }
 
 impl Task {
@@ -41,6 +42,7 @@ impl Task {
         path: String,
         starting_status: TaskStatus,
         enqueue_at: Option<DateTime<Local>>,
+        dependencies: Vec<usize>,
     ) -> Task {
         Task {
             id: 0,
@@ -54,6 +56,7 @@ impl Task {
             stderr: None,
             start: None,
             end: None,
+            dependencies,
         }
     }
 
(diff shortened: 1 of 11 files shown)


shared/task.rs
```diff
@@ -1,89 +1,404 @@
...
+                    .iter()
+                    .flat_map(|id| state.get_task(*id))
+                    .filter(|task| !task.is_done())
+                    .collect();
+
+                if !dep_not_done.is_empty() {
+                    let failed = dep_not_done.iter().find(|task| task.is_errored());
+
+                    if let Some(failed_task) = failed {
+                        Startability::DependenciesFailure(failed_task.id)
```
Nukesor, line 262: A tasks with failed dependencies should fail hard.
As far as I understand, right now those tasks will still be in internal `Waiting` state, but displayed as failed to the outside?

shared/task.rs
```diff
@@ -1,89 +1,404 @@
...
+                    Startability::Ready
+                }
+            }
+            _ => Startability::Unknown,
+        }
+    }
+
+    pub fn general_state(&self, state: &State) -> GeneralState {
+        match &self.state {
+            TaskState::Waiting { .. } => match self.start_info(state) {
```
Nukesor, line 276: Without DependenciesFailure and Unknown (which IMO is really bad practice) the whole start_info stuff can be removed.
The only place this should be used in, is in the TaskHandler, since that's the only place where we have to check whether a task can be started or not.

Nukesor:
I merged your other PR, added output handling and tinkered a little around to create a smaller refactoring inspired by your PR, to simplify handling of task results.

Feel free to check it out:
https://github.com/Nukesor/pueue/pull/87

PR #88, opened 2020-04-30
Process output no longer stored in memory

### LINKED ISSUE #82: [BUG] excessive memory and CPU usage with large outputs

Nukesor:
Resolves #82


client/output.rs
```diff
@@ -171,48 +171,51 @@ pub fn print_logs(tasks: BTreeMap<usize, Task>, cli_command: &SubCommand) {
...
     if json {
-        println!("{}", serde_json::to_string(&tasks).unwrap());
+        println!("{}", serde_json::to_string(&task_logs).unwrap());
         return;
     }
 
-    if task_ids.is_empty() && tasks.is_empty() {
-        println!("There are no finished tasks");
+    if task_ids.is_empty() && task_logs.is_empty() {
+        println!("There are no finished task_logs");
```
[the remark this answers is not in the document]
Nukesor: Nice catch. Search and replace did a little bit too much over here.

CHANGELOG.md
```diff
@@ -1,4 +1,18 @@
...
+    Tasks with failed dependencies will fail as well.
+- New state `FailedToStart`. Used if the process cannot be started
+- New state `DependencyFailed`. Used if any dependency of a task fails.
+
+**Improvements:**
+- Process log output is no longer stored in memory. This significantly reduced RAM usage for large log outputs.
+- Process log output is compressed in-memory on read from disk. This leads to reduced bandwidth and RAM usage.
+
+**Changes:**
+- Pueue no longer stores log output in it's backup files
```
sourcefrog, line 13: It's just "it's" for the possessive, analogous to "his" or "hers".

client/output.rs
```diff
@@ -232,27 +239,40 @@ pub fn print_log(task: &Task) {
...
-            println!(
-                "{}",
-                style("Std_out: ")
-                    .with(Color::Green)
-                    .attribute(Attribute::Bold)
-            );
-            println!("{}", stdout);
+    if !task_log.stdout.is_empty() {
+        if let Err(err) = print_task_output(&task_log, true) {
+            println!("Error while parsing stdout: {}", err);
```
sourcefrog, line 244: Should these be "parsing" or "printing"?

client/output.rs
```diff
@@ -232,27 +239,40 @@ pub fn print_log(task: &Task) {
...
-            println!("{}", stderr);
-        }
+    if !task_log.stderr.is_empty() {
+        if let Err(err) = print_task_output(&task_log, false) {
+            println!("Error while parsing stderr: {}", err);
+        };
     }
 }
+
+/// Pritn the stdout of a finished process
```
sourcefrog, line 255: *Print

client/output.rs
```diff
@@ -232,27 +239,40 @@ pub fn print_log(task: &Task) {
...
+/// The logs are compressed using Brotli and then encoded to Base64
```
sourcefrog, line 256: Is it Brotli or Snappy?

CHANGELOG.md
```diff
@@ -1,4 +1,18 @@
-# v0.3.0
+# v0.4.0
```
sourcefrog, line 1: It seems like new clients will be confused by trying to talk to old servers, with this protocol change. Do you want to add a warning to restart them? Or even some inline marker in the connection?

CHANGELOG.md
sourcefrog: And (such a tiny nit) some of the lines are missing periods.

client/output.rs
Nukesor: Snappy. Forgot to change the comment.

Brotli was way to slow

client/output.rs
Nukesor: That should be parsing. This only occurs if some kind of error during deserialization occurred.

daemon/log.rs
```diff
@@ -63,3 +65,27 @@ pub fn clean_log_handles(task_id: usize, settings: &Settings) {
         );
     };
 }
+
+/// Return stdout and stderr of a finished process
+/// Everything is compressed using Brotli and then encoded to Base64
```
sourcefrog, line 70: Snappy here too?

daemon/log.rs
```diff
@@ -63,3 +65,27 @@ pub fn clean_log_handles(task_id: usize, settings: &Settings) {
...
+pub fn read_and_compress_log_files(
+    task_id: usize,
+    settings: &Settings,
+) -> Result<(Vec<u8>, Vec<u8>)> {
+    let (mut stdout_handle, mut stderr_handle) = match get_log_file_handles(task_id, settings) {
+        Ok((stdout, stderr)) => (stdout, stderr),
+        Err(err) => return Err(anyhow!(format!("Error while opening the output files: {}", err))),
```
sourcefrog, line 77: I think you can use [`bail!()`](https://docs.rs/anyhow/1.0.28/anyhow/macro.bail.html) here

daemon/log.rs
sourcefrog, line 70: I haven't followed the logic really closely, but is it actually necessary to base64-encode if you're sending it over bincode? Isn't that binary-safe?

Nukesor:
Thanks for the extensive reviews :) I really appreciate your help!

I'm in the middle of making the whole "logs sent via socket" thingy optional. The new default will be to read the files directly from disk for clients.

I agree that exposing the service adds a strong and powerful attack vector for RCE.
That's why I restricted any communication to loopback for now.

I'm waiting for a fool-proof and easy-to-use low-level tcp crypto-library to pop up.
SSH port forwarding can be used for now as a workaround, but I would love to get this supported natively. (I tried to look around for combining the current solution and SSH inside of Rust, but it doesn't look like there are good libraries out there for this problem yet).

However, I won't implement this, until I'm VERY certain that the underlying crypto is safe (Which will probably take quite some time). It's something I would like to see added to the project and I like to keep it in mind, but I'll try to make it work without it.

Hence, the new feature to read log files locally. You were absolutely right about this :)

sourcefrog:
Thanks! It's been fun working with you on this.

I think the popular thing these days for such situations is mTLS, Mutual TLS. It does still have a challenge of generating and managing keys and access, which needs the user to have some infrastructure and beyond just the library.

Some background reading, but not drop-in solutions for Rust:

* https://cloud.google.com/security/encryption-in-transit/application-layer-transport-security - not literally TLS but similar
* https://smallstep.com/ -- apparently pretty lightweight key management infrastructure - I haven't used it

daemon/log.rs
Nukesor: That was an old comment as well^^
The base64 encode was needed before migrating to bincode, since binary data isn't natively supported in JSON.

daemon/log.rs
Nukesor: Nice. Didn't know about that one!

CHANGELOG.md
Nukesor: Added

Nukesor:
Ok. I think this can finally be merged into master.

I'll probably run the new master version for a few days on my machines before releasing a new version, though.

PR #94, opened 2020-05-30
Fix formatting of log header

sourcefrog:
Fixes Nukesor/pueue/#93


Nukesor:
Thanks :D

I'm using it every few days and somehow didn't notice :sweat_smile:

Issue #95, opened 2020-06-01
Defaults for Follow


JP-Ellis:
It would be good if the `follow` subcommand was a little smarter if no tasks are specified.  In particular:

- If there are no tasks running, `pueue follow` should simply say that there are no tasks running.
- If there is only one task running, `pueue follow` should follow that running task
- If there are multiple tasks running, `pueue follow` could either:
  - List which tasks are running and warn the user a task ID has to be specified; or,
  - Follow the last task

In the last case, I prefer the first option as it doesn't assume anything unreasonable.

In all case, `pueue follow <task-id>` should remain unchanged.

If you would like this added, I'm happy to work on a PR request.

Nukesor:
I like it! This should definitely improve the ergonomics of the sub command.

Feel free to start hacking, I would love to see this added to the project :+1:

Nukesor:
Hey @JP-Ellis,

I just wanted to ask, when you're planning to build this feature :)
I'm thinking about creating a new release, but I would like to wait for your feature, so it can be included in 0.6.0.

If you need some time, I'll push a patch release for the new features :)

Anyhow, no pressure, just take your time :)

JP-Ellis:
I was hoping to have a go at this this weekend.

PR #98, opened 2020-06-08
Implement Defaults for Follow

JP-Ellis:
This implements #95.

As this requires communication with the daemon in order to know which tasks are running, the Client code was slightly refactored in order to allow the connection to be re-used.

Also moved the `get_message_from_opt` into Client as the conversion from Opt to Message requires a possible query to the daemon (to get the running tasks).


client/output.rs
```diff
@@ -185,7 +185,8 @@ pub fn print_logs(
 ) {
     let (json, task_ids) = match cli_command {
         SubCommand::Log { json, task_ids } => (*json, task_ids.clone()),
-        SubCommand::Follow { task_id, .. } => (false, vec![*task_id]),
+        // TODO: Check that SubCommand::Follow is also unreachable.
```
Nukesor, line 188: What exactly do you mean by that?

client/output.rs
```diff
@@ -354,9 +355,13 @@ pub fn print_remote_task_output(task_log: &TaskLogMessage, stdout: bool) -> Resu
     Ok(())
 }
 
-/// Print the log ouput of finished tasks.
-/// Either print the logs of every task
-/// or only print the logs of the specified tasks.
+/// Follow the log ouput of running task.
+///
+/// If no task is specified, this wil check for the following cases:
```
Nukesor, line 360:
```suggestion
/// If no task is specified, this will check for the following cases:
```

client/output.rs
```diff
@@ -354,9 +355,13 @@ pub fn print_remote_task_output(task_log: &TaskLogMessage, stdout: bool) -> Resu
     Ok(())
 }
 
-/// Print the log ouput of finished tasks.
-/// Either print the logs of every task
-/// or only print the logs of the specified tasks.
```
Nukesor, line 359: Good old copy-paste comment :sweat: 
Thanks for fixing this.

JP-Ellis:
With regards to the client refactoring and moving `get_message_from_opt` into the client code, I agree it would be nice to refactor it some more, but I didn't want to do too invasive changes for this pull request.

I think there's two ways of doing this:

1. Have the client handle the conversion from `Opt` to `Message`.  This is the closest to what was in the code hence what I did.
2. Implement an `into_message` method for `Opt`.

With regards to refactoring the Client code, I agree that it could be improved a little, though I'm definitely no expert.  I just feel like it does several things at once and it would make it perhaps a little easier if we separated out the following:

1. Instantiating a new Client
2. Parsing of command line argument to Message
3. Sending message
4. Receiving / Handling a message

The one that sticks out from all the rest is definitely the second point, hence why I think implementing an `into_message` method my best?  Perhaps a discussion of refactoring the Client should be kept to a separate issue though.

client/output.rs
JP-Ellis: Well, from what I could tell, this function is only ever called when the logs are being printed out and thus the `Follow` sub-command should never appear.  I left it commented just in case I misunderstood the logic though.

Nukesor: Nope, thats correct :)
This path should only be entered with `Log`

Nukesor:
Well, everything looks alright and I'm going to merge this PR :)
I'll remove the small TODO comment while I'm at it ;)

Thanks for contributing!

Nukesor:
You're right, the client refactoring is something for another time and PR.

The biggest problem I see within the client right now is, that there's a lot of Stuff that can be handled in a similar fashion, while a few commands need a LOT of special handling.

It would be super nice to get a nicer and more generic way to handle requests, which would result in a better separation of client setup, communication and response handling / multiple request handling.

Well, it works for now, but I'm sure that this will make problems at some point in the future. At the latest, if somebody decides to build a tui-rs interface for Pueue :D

PR #120, opened 2020-10-18
Add warnings when removing/killing tasks

### LINKED ISSUE #111: Warning messages for removing/killing tasks
A warning message for removing a bunch of tasks may be useful:

```bash
$ pueue remove 5-8
you are trying to remove:
task5, task6, task7, task8

Are you sure to remove tasks?[Y/n]
```

_Originally posted by @dhra in https://github.com/Nukesor/pueue/issues/108#issuecomment-669846883_

This should be entirely optional and configurable via a flag in the configuration file!

Any help is appreciated, since this is something I won't implement myself. Tbh, I just don't need it.

KaindlJulian:
Like how it is described in #111 this adds completely optional user confirmations for removing and killing tasks.

The config file flag is `print_remove_warnings` which is false by default. (not too fond of the naming tbh ^^, maybe someone can suggest a better one)

With `print_remove_warnings: true`:
image

image

Resolves #111

### DIFF
--- a/client/client.rs
+++ b/client/client.rs
@@ -185,13 +185,13 @@ impl Client {
 
     /// Prints a warning and prompt for given action and tasks.
     /// Returns `Ok(())` if the action was confirmed.
-    fn handle_user_confirmation(&self, action: &str, task_ids: &Vec<usize>) -> Result<()> {
+    fn handle_user_confirmation(&self, action: &str, task_ids: &[usize]) -> Result<()> {
         // printing warning and prompt
         println!(
             "You are trying to {}: {}",
             action,
             task_ids
-                .into_iter()
+                .iter()
                 .map(|t| format!("task{}", t.to_string()))
                 .collect::<Vec<String>>()
                 .join(", ")
@@ -202,7 +202,7 @@ impl Client {
         io::stdout().flush().unwrap();
         io::stdin().read_line(&mut input)?;
 
-        if input.starts_with("n") || input.starts_with("N") {
+        if input.starts_with('n') || input.starts_with('N') {
             bail!("User did not confirm");
         }
 
--- a/client/client.rs
+++ b/client/client.rs
@@ -196,17 +196,26 @@ impl Client {
                 .collect::<Vec<String>>()
                 .join(", ")
         );
-        print!("\nDo you want to continue [Y/n]: ");
 
         let mut input = String::new();
-        io::stdout().flush().unwrap();
-        io::stdin().read_line(&mut input)?;
 
-        if input.starts_with('n') || input.starts_with('N') {
-            bail!("User did not confirm");
+        loop {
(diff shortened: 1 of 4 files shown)


client/client.rs
```diff
@@ -183,6 +183,32 @@ impl Client {
...
+                .collect::<Vec<String>>()
+                .join(", ")
+        );
+        print!("\nDo you want to continue [Y/n]: ");
+
+        let mut input = String::new();
+        io::stdout().flush().unwrap();
+        io::stdin().read_line(&mut input)?;
+
+        if input.starts_with('n') || input.starts_with('N') {
```
Nukesor, line 205: This might be a good place to use a `loop`.
It's probably a good idea to check whether the user entered either `Y/y` or `N/n`. Right now it is possible that the user enters any other character than `N/n` and the input will be accepted as if it was an `y`.

Nukesor (changes requested):
Nice! I like the generic function for user confirmations :)

Nukesor:
It would also be awesome, if you could add the new option to the [readme config section](https://github.com/nukesor/pueue#configs).

While you're at it, you can also add your change to the `CHANGELOG.md`. Don't forget to mention yourself :) 

I suggest you rebase/merge master beforehand, since I just added/changed stuff in both those files :)

client/client.rs
```diff
@@ -183,6 +183,32 @@ impl Client {
...
+            bail!("User did not confirm");
```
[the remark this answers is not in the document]
KaindlJulian: yea the calls are a lot cleaner now :+1:

client/client.rs
KaindlJulian: Should work as intended now, I totally forgot to about the other options :smile:

client/client.rs
```diff
@@ -277,6 +321,9 @@ impl Client {
                 all,
                 children,
             } => {
+                if self.settings.client.print_remove_warnings {
```
Nukesor, line 324: One last thing :D

`print_remove_warnings` is probably not the best name for this feature.
It's not only a print, but also a question and it's shown on `kill` and `remove`.

I'm struggling to come up with a good name as well, though. Stuff like that is always super hard.
I came up with something like `show_confirmation_questions`.

That should hopefully clarify what this option is used for, without having to read the `README`.

Sorry for only mentioning this now. It didn't strike me in the first review :D

Nukesor (changes requested):

client/client.rs
KaindlJulian: No worries :D best I could think of was `require_confirmations` or so. But I ended up using the one you suggested because it contains more information.

Nukesor (approved):

Nukesor:
Looks good to me!
Thanks for coding this feature :)

PR #124, opened 2020-10-26
Fix typos in Readme and help

Sh3Rm4n:
Sorry for the low-effort PR 😅 

Here are some typos I find while checking out this tool :)


Nukesor:
Thanks a lot, very much appreciated :)

And don't worry, it's never low-effort as long as it's useful!

PR #126, opened 2020-10-26
fix zsh completion hang

### LINKED ISSUE #57: Automatic ZSH Completion script generation breaks
The script does compile, but completion doesn't work however.

Someone with a little bit of zsh completion script knowledge probably needs to look at that script.

Run `./build_completions.sh` and check the _pueue file in the utils/completions directory

ahkrr:
This PR fixes the issue of the zsh completion script hanging when executing.

The doc-comment in the cli will be part of the completion string which would result in this scenario
    "This command works like \`tail -f\`"
At the time of evaluation zsh will call tail -f, which will then do its job 

Removing the \` \` will prevent this from happening
    "This command works like tail -f"

This will fix #57

### DIFF
--- a/client/cli.rs
+++ b/client/cli.rs
@@ -252,7 +252,7 @@ pub enum SubCommand {
     },
 
     /// Follow the output of a currently running task.
-    /// This command works like `tail -f`.
+    /// This command works like tail -f.
     Follow {
         /// The id of the task you want to watch.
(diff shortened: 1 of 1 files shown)


Nukesor:
Daaamn,

I thought, that I managed to fix all those unescaped symbols on my last revisit and something else is fundamentally broken...

Thanks a lot!

PR #127, opened 2020-10-26
Add times to callback parameters

soruh:
This adds the start, end and enqueue time to the callback parameters or the empty string if they are not present.

### DIFF
--- a/daemon/task_handler.rs
+++ b/daemon/task_handler.rs
@@ -741,6 +741,15 @@ impl TaskHandler {
         parameters.insert("command", task.command.clone());
         parameters.insert("path", task.path.clone());
         parameters.insert("result", task.result.clone().unwrap().to_string());
+
+        let print_time = |time: Option<DateTime<Local>>| {
+            time.map(|time| time.timestamp().to_string())
+                .unwrap_or_else(String::new)
+        };
+        parameters.insert("enqueue", print_time(task.enqueue_at));
+        parameters.insert("start", print_time(task.start));
+        parameters.insert("end", print_time(task.end));
+
         if let Some(group) = &task.group {
             parameters.insert("group", group.clone());
         } else {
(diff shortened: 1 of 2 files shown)


soruh:
The CI seems to have failed due to a DNS issue on the runner.
I would probably make sense to rerun it when that issue is resolved.

daemon/task_handler.rs
```diff
@@ -741,6 +741,15 @@ impl TaskHandler {
         parameters.insert("command", task.command.clone());
         parameters.insert("path", task.path.clone());
         parameters.insert("result", task.result.clone().unwrap().to_string());
+
+        let print_time = |time: Option<DateTime<Local>>| {
+            time.map(|time| time.timestamp().to_string())
+                .unwrap_or_else(String::new)
```
Nukesor, line 747: Hmm.
I'm not sure if a timestamp is the best solution.
It might be a good solution to let the user decide how they want to format stuff.
It's pretty unergonomic, though.

Let me think about this for a little bit :D

soruh: I thought it would be the most versatile, you can convert it with one inline call to `date` after all. I do agree though that this is something that should be carefully decided because changing after the fact would break every single callback hook that uses it.

soruh: For what it's worth it's also easier to calculate the runtime this way, since you can just subtract the timestamps instead of having to convert both dates to timestamps first.

Nukesor: Ok. After a good night I agree with you :D
Let's use timestamps.

Nukesor:
Code looks good to me.
I have a little request, though. Would you mind to create a second PR against the `wiki` branch of the project and update the callback section of the "Miscellaneous" page?

Also go ahead and add yourself to the changelog :)

There's a no preset for the next version, so just add this:
```
## [0.8.1] -

### Added
```

soruh:
Great, thank you. I'll do that in a bit when I'm back at my computer.

Nukesor:
Perfect.
Thanks a lot for you contribution :+1:

soruh:
Great. Happy to be be able to contribute.

PR #128, opened 2020-10-27
Document time parameters for callbacks

soruh:
- Document time parameters for callbacks
- Update callback example to include runtime
- Add example for external callback hook


Nukesor:
LGTM!

PR #149, opened 2020-12-13
Add Alpine Linux to list of distributions

jirutka:
Note: Alpine 3.13 has not been released yet, it should be till the end of this year.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ Use your system's package manager.
 This will usually deploy service files and completions automatically.  
 Pueue has been packaged for:
 
+- Alpine Linux (since 3.13)
 - Arch Linux's AUR: e.g. `yay -S pueue`.
 - NixOS
 - Homebrew


Nukesor:
I'm really happy to hear that Pueue managed to get packaged for Alpine :)!

Do you think we should already add this to the Readme or rather wait until Alpine 3.13 get's officially released?

jirutka:
> I'm really happy to hear that Pueue managed to get packaged for Alpine :)!

Creating an [APKBUILD](https://github.com/alpinelinux/aports/blob/master/community/pueue/APKBUILD) for pueue was easier for me than remembering its name (still struggling). ;)

> Do you think we should already add this to the Readme or rather wait until Alpine 3.13 get's officially released?

It will be probably better to wait.

Nukesor:
I just looked around a little and found this super helpful website:
https://repology.org/project/pueue/versions

This automatically creates very nice packaging info for pretty much every major distro/OS out there:

<a href="https://repology.org/project/pueue/versions">Packaging status</a>

Nukesor:
I'll close this PR in favor of the generic solution :)

Anyway, thanks for your contribution and the Alpine packaging!

PR #174, opened 2021-01-30
Minor option docstring cleanup

hartzell:
This was a bit confusing until I skimmed the code.  Clarifying for posterity.


Nukesor:
thanks!

PR #195, opened 2021-04-29
Change macos target to apple target

Nukesor:
Related pueue-lib PR:
https://github.com/Nukesor/pueue-lib/pull/2


Nukesor:
@althiometer Not sure what's the problem. CI is failing, but I don't understand why.
Could you go ahead and help out debugging?

Nukesor:
Ah. got it. `target_vendor` instead of `target_os`. Sorry for the ping.

althiometer:
Yep, thanks for this! This is gonna be really cool to use on the other apple platforms without some funky patching in the makefiles!

Issue #191, opened 2021-04-10
Allow `~` in Pueue config paths


dadav:
**Describe the bug**
The documentation says you can use e.g. `$USER` or `$HOME` (shell in general) in the configuration file (https://github.com/Nukesor/pueue/wiki/Configuration). At least it seams to me like that...But this results in a literal `$HOME` directory.

**To Reproduce**
Use `pueue_directory: $HOME/.local/share/pueue` in your config.

**Expected behavior**
The daemon should resolve the path the the actual location.

**Additional context**
OS: Archlinux
pueue daemon: 0.12.1

Nukesor:
Those are example placeholders that should be replaced by the actual thing.
I see how this can be confusing.

I'm open to suggestions on how to clarify this :)

dadav:
Wouldn't it be nice to actually support variables? :)

Nukesor:
On Unix, at least the home `~` notation is supported :)

dadav:
> 
> 
> On Unix, at least the home `~` notation is supported :)

Ohhh, okay, thats all I need!!

dadav:
Hi, I just realized that there is a literal ~ in my homedir :D

Nukesor:
Wow. Yep, I'm able to reproduce that behavior.
Well, I guess `~` isn't supported after all :D .

Ok, there are a few different possible approaches:

1. First solution:

Resolve anything like `~` or similar stuff during deserialization of the configuration file.
This comes with its own caveats.

The daemon manipulates and overwrites the configuration file regularily (for instance if you add a group).
As a result, you would also have to revert the original resolving on serialization. \
But how do you know whether there originally was a `~` or a `/home/your_username/`?

I don't like this one too much.

2. Second solution:

Don't touch any strings during deserialization/serialization.
Instead all resolving/parsing would have to be done in getter functions and make all paths non-public struct members.

This is most likely the cleanest solution, but will require quite a bit of refactoring, since those paths are passed around pretty much everywhere.


If you feel like this is something you must have, feel free to start coding :D.

However, it's nothing I'll implement myself, since i haven't touched my Pueue config file in ages :D

dadav:
I never wrote a single line in rust, but I'll try :D Can't be that hard, right? *fingers crossed*

dadav:
[...]

PR #197, opened 2021-05-04
Use getter methods which actually expand tilde

dadav:
See #191

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -523,13 +523,34 @@ dependencies = [
 ]
 
 [[package]]
+name = "dirs-next"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


dadav:
Is this ok? Anything left for me to do? :)

Nukesor:
Ignore the failing CI, this is due to the new pueue-lib version not being released yet

Nukesor:
Now that I think about it, let's try something.

I just created a new branch `v0.13`, which uses the current `pueue-lib` master.
This should allow us to develop both branches concurrently without having to wait for any new releases.

Could you try and change the merge target from `master` to `v0.13`?
If this works, it could make things a lot easier :)

dadav:
Done

dadav:
Thanks for this awesome tool, I love it :)

PR #203, opened 2021-05-19
Pueue signal

### LINKED ISSUE #202: Having an option to send SIGTERM instead of SIGKILL to a program.
**Is your feature request related to a problem? Please describe.**
First thanks for building this fantastic library! I was really excited to include `pueue` into my daily command-line workflow the moment I found it. 

Basically, I have some programs react with SIGTERM to do a graceful shutdown. However, it seems like `pueue` will send SIGKILL to the program for`pueue kill`, which is not capturable. 

**Describe the solution you'd like**
Maybe add a `-s` parameter to `pueue kill` to allow us select the signal?

Nukesor:
Resolves #202 

This should work. The signal parameter isn't case insensitive yet, so you have to write `SigInt` etc..
I'll check on how to fix this as soon as I get back to it.

@dovahcrow Could you check it and see if everything works as expected :)?

`pueue add 'sleep 10' && pueue add sleep 20 && sleep 1 && pueue signal SigStop 0 1`

### DIFF
--- a/client/cli.rs
+++ b/client/cli.rs
@@ -5,6 +5,8 @@ use chrono::Duration;
 use chrono_english::*;
 use clap::Clap;
 
+use pueue_lib::network::message::Signal;
+
 #[derive(Clap, Debug)]
 pub enum SubCommand {
     /// Enqueue a task for execution.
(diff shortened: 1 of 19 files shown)


client/cli.rs
```diff
@@ -224,6 +226,25 @@ pub enum SubCommand {
         input: String,
     },
 
+    /// Send a UNIX signal to one or multiple tasks.
+    /// DISCLAIMER: This bypasses Pueue's process handling logic!
+    ///     You might enter weird invalid states, use at your own descretion.
+    #[cfg(not(target_os = "windows"))]
+    Signal {
```
dovahcrow, line 233: It is easy to implement the `-g` flag? i.e. send the signal to a group.

Nukesor: I.e. sending the signal to all running tasks of a specific group?

dovahcrow: yep

Nukesor: After some consideration, I went back and decided to move the signal handling to the `--signal, -s` flag to kill.

The logic is basically the same and we only need very few adjustments.
The help text on the flag should also clarify things and warn the user about their actions.

dovahcrow: Thanks! Yeah making the flag only work under POSIX systems also works.

PR #206, opened 2021-05-20
fix children process handling issue under linux

### LINKED ISSUE #205: [BUG] Kill shell should first gather the children processes
**Describe the bug**

https://github.com/Nukesor/pueue/blob/v0.13/daemon/platform/linux/process_helper.rs#L108 seems not correct to me.

If the parent process is killed first at line 108, then line 112 will always return empty because the parent process is gone then the children are now having pid1 as the parent.

dovahcrow:
fixes #205 205


dovahcrow:
Oops wrong target branch

PR #207, opened 2021-05-20
fix children process handling issue under linux

### LINKED ISSUE #205: [BUG] Kill shell should first gather the children processes

dovahcrow:
fixes #205


Nukesor:
Wow. That one's actually quite critical.
I remember stumbling over some detached processes a while ago, but I couldn't find the cause for this as I didn't know how to reproduce the issue.

You might just have fixed that!. Thanks!

Nukesor:
Ah no, that couldn't have been it, since this didn't lead to exiting processes before.
The logic in the actual killing of tasks were handled correctly.

PR #270, opened 2021-12-10
Add --lines to follow

drewkett:
This adds a `--lines` option to the follow command. I also added a more efficient implementation of read_last_lines to `pueue-lib` (compared to the simple one in there now). I branched off `pueue-lib` off of `0.18.1` because it was the quickest way for me to test it since I wasn't sure if you've made any breaking changes in the work towards the new version. I'll make a separate PR for `pueue-lib` for visibility (Nukesor/pueue-lib#18). If you're interested in merging these PR's, let me know the preferred method of getting it incorporated with respect to which branch to use as the base and how to coordinate `pueue` and `pueue-lib` changes. I also see you're working on a new version so maybe this should just target that release? Let me know

### DIFF
--- a/client/display/follow.rs
+++ b/client/display/follow.rs
@@ -3,7 +3,7 @@ use std::path::Path;
 use std::thread::sleep;
 use std::time::Duration;
 
-use pueue_lib::log::{get_log_file_handles, get_log_paths};
+use pueue_lib::log::{get_log_file_handles, get_log_paths, seek_to_last_lines};
 
 /// Follow the log ouput of running task.
 ///
@@ -12,7 +12,12 @@ use pueue_lib::log::{get_log_file_handles, get_log_paths};
 /// - No running task: Print an error that there are no running tasks
 /// - Single running task: Follow the output of that task
 /// - Multiple running tasks: Print out the list of possible tasks to follow.
-pub fn follow_local_task_logs(pueue_directory: &Path, task_id: usize, stderr: bool) {
+pub fn follow_local_task_logs(
+    pueue_directory: &Path,
+    task_id: usize,
+    stderr: bool,
+    lines: Option<usize>,
+) {
     let (stdout_handle, stderr_handle) = match get_log_file_handles(task_id, pueue_directory) {
         Ok((stdout, stderr)) => (stdout, stderr),
         Err(err) => {
@@ -28,6 +33,15 @@ pub fn follow_local_task_logs(pueue_directory: &Path, task_id: usize, stderr: bo
     // Stdout handler to directly write log file output to io::stdout
     // without having to load anything into memory.
     let mut stdout = io::stdout();
+
+    // If lines is passed as an option, seek the output file handle to the start of
+    // the line corresponding to the `lines` number of lines from the end of the file.
+    // The loop following this section will copy those lines to stdout
+    if let Some(lines) = lines {
+        if let Err(err) = seek_to_last_lines(&mut handle, lines) {
+            println!("Error seeking to last lines from log: {}", err);
+        }
+    }
     loop {
         // Check whether the file still exists. Exit if it doesn't.
         if !handle_path.exists() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - When calling `pueue` without a subcommand, the `status` command will be called by default [#247](https://github.com/Nukesor/pueue/issues/247).
 - Add the `--group` parameter to the `pueue clean` command [#248](https://github.com/Nukesor/pueue/issues/248)
 - Add `stdout_path` and `stderr_path` as template parameters for callbacks [#269](https://github.com/Nukesor/issues/269)
+- Add `--lines` parameter to `pueue follow` to only show specified number of lines from stdout before following [#270](https://github.com/Nukesor/pueue/issues/270)
 
 ### Changed
 
--- a/client/cli.rs
+++ b/client/cli.rs
@@ -329,6 +329,10 @@ pub enum SubCommand {
         /// Show stderr instead of stdout.
         #[clap(short, long)]
         err: bool,
+
+        /// Only print the last X lines of the output before following
+        #[clap(short, long)]
+        lines: Option<usize>,
     },
 
     /// Wait until tasks are finished. This can be quite useful for scripting.
--- a/client/client.rs
+++ b/client/client.rs
@@ -215,7 +215,11 @@ impl Client {
                 Ok(true)
             }
 
-            SubCommand::Follow { task_id, err } => {
+            SubCommand::Follow {
+                task_id,
+                err,
+                lines,
+            } => {
                 // Simple log output follows for local logs don't need any communication with the daemon.
                 // Thereby we handle this separately over here.
                 if self.settings.client.read_local_logs {
@@ -224,6 +228,7 @@ impl Client {
                         &self.settings.shared.pueue_directory(),
                         task_id,
                         *err,
+                        *lines,
                     )
                     .await?;
                     return Ok(true);
@@ -499,10 +504,15 @@ impl Client {
                 };
                 Ok(Message::Log(message))
             }
-            SubCommand::Follow { task_id, err } => {
+            SubCommand::Follow {
+                task_id,
+                err,
+                lines,
+            } => {
                 let message = StreamRequestMessage {
                     task_id: *task_id,
                     err: *err,
+                    lines: *lines,
                 };
                 Ok(Message::StreamRequest(message))
             }
--- a/client/commands/local_follow.rs
+++ b/client/commands/local_follow.rs
@@ -12,6 +12,7 @@ pub async fn local_follow(
     pueue_directory: &Path,
     task_id: &Option<usize>,
     err: bool,
+    lines: Option<usize>,
 ) -> Result<()> {
     // The user can specify the id of the task they want to follow
     // If the id isn't specified and there's only a single running task, this task will be used.
@@ -46,7 +47,7 @@ pub async fn local_follow(
         }
     };
 
-    follow_local_task_logs(pueue_directory, task_id, err);
+    follow_local_task_logs(pueue_directory, task_id, err, lines);
 
     Ok(())
 }
--- a/daemon/network/follow_log.rs
+++ b/daemon/network/follow_log.rs
@@ -73,6 +73,14 @@ pub async fn handle_follow(
     let (out_path, err_path) = get_log_paths(task_id, pueue_directory);
     let handle_path = if message.err { err_path } else { out_path };
 
+    // If lines is passed as an option, seek the output file handle to the start of
+    // the line corresponding to the `lines` number of lines from the end of the file.
+    // The loop following this section will copy those lines to stdout
+    if let Some(lines) = message.lines {
+        if let Err(err) = seek_to_last_lines(&mut handle, lines) {
+            println!("Error seeking to last lines from log: {}", err);
+        }
+    }
     loop {
         // Check whether the file still exists. Exit if it doesn't.
         if !handle_path.exists() {
--- a/lib/src/network/message.rs
+++ b/lib/src/network/message.rs
@@ -209,6 +209,7 @@ pub enum Shutdown {
 pub struct StreamRequestMessage {
     pub task_id: Option<usize>,
     pub err: bool,
+    pub lines: Option<usize>,
 }
 
 /// Request logs for specific tasks.


Nukesor:
I decided to pull the `pueue-lib` project back into the main repository!

Could you merge rebase onto the current `development` HEAD your changes in pueue-lib into this MR?
This will should make colaburation much easier :)

drewkett:
That definitely simplifies this change. This should bow be up to date on the development branch. It compiles and tests fine but haven't actually tested the functionality again with the move to the development branch.

lib/src/log.rs
```diff
@@ -129,22 +130,84 @@ pub fn reset_task_log_directory(path: &Path) -> Result<(), Error> {
     Ok(())
 }
 
+pub fn read_last_lines_as_byte_deque(file: &mut File, lines: usize) -> Result<VecDeque<u8>, Error> {
+    // The integer types were chosen to avoid overflows when converting between signed and unsigned
+    // integers. u16 is used here so that we can support 16bit platforms with simple integer casts
+    // (FIXME Not sure if pueue even is supported on any 16bit platforms)
+    let chunk_size = 4096u16;
```
Nukesor, line 137: Pueue isn't used on 16bit platforms. At least I'm not aware of it.
Anyhow, u64 should still be supported u16 systems anyway. There are also u128 on 64-bit systems. The number will then just occupy four memory addresses instead of a single one.

I think it's best to switch this to `u64`. This will remove a lot of number casting and error handling.

lib/src/log.rs
```diff
@@ -129,22 +130,84 @@ pub fn reset_task_log_directory(path: &Path) -> Result<(), Error> {
     Ok(())
 }
 
+pub fn read_last_lines_as_byte_deque(file: &mut File, lines: usize) -> Result<VecDeque<u8>, Error> {
```
Nukesor, line 133: Missing docstring.
This function contains a quite complex algorithm. This should be well documented.

lib/src/log.rs
```diff
@@ -129,22 +130,84 @@ pub fn reset_task_log_directory(path: &Path) -> Result<(), Error> {
...
+    // The integer types were chosen to avoid overflows when converting between signed and unsigned
+    // integers. u16 is used here so that we can support 16bit platforms with simple integer casts
+    // (FIXME Not sure if pueue even is supported on any 16bit platforms)
+    let chunk_size = 4096u16;
+    let mut buf = vec![0u8; chunk_size as usize];
+    let size = match file.seek(io::SeekFrom::End(0)) {
+        Ok(p) => p,
```
Nukesor, line 140: Please don't use single-character variable names.
Especially if they're rather confusing. Please rename `p` to `size` or `seeked_bytes`.

lib/src/log.rs
```diff
@@ -129,22 +130,84 @@ pub fn reset_task_log_directory(path: &Path) -> Result<(), Error> {
...
+        Err(err) => {
+            return Err(Error::LogSeek(format!(
+                "Error while seeking log file : {}",
+                err
+            )));
+        }
+    };
+    let mut position = size;
```
Nukesor, line 148: Please add comments to runtime variables.
Or make the variable names more explicit, such as `cursor_position`.

lib/src/log.rs
```diff
@@ -129,22 +130,84 @@ pub fn reset_task_log_directory(path: &Path) -> Result<(), Error> {
...
+    let mut text = std::collections::VecDeque::new();
```
Nukesor, line 149: Same here.
`text` could be everything. `last_lines` would be a much better fit.

lib/src/log.rs
```diff
@@ -129,22 +130,84 @@ pub fn reset_task_log_directory(path: &Path) -> Result<(), Error> {
...
+        }
+        // Safe to cast chunk_size from u16 to u64
+        let next_chunk_size = if position > chunk_size as u64 {
+            chunk_size
+        } else {
+            // Safe to cast position to u16 since it is <= chunk_size which is u16
+            position as u16
+        };
+        // Safe to cast next_chunk_size from u16 to i64
+        position = match file.seek(io::SeekFrom::Current(-(next_chunk_size as i64))) {
```
Nukesor, line 163: This is very equal to the current behavior of https://doc.rust-lang.org/std/io/struct.BufReader.html.

I checked whether there already is a crate for this and it turns out there is!
https://docs.rs/rev_buf_reader/latest/rev_buf_reader/

The project is rather small, but it is very well documented and most importantly properly tested.

To be honest, I don't feel too comfortable to include such a finnicky algorithm, if there's a proper and tested implementation out there. Just like the official BufReader it can also read lines in reverse.

I made a small playground project:

Cargo.toml:
```
[package]
name = "rofl"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
rev_buf_reader = "0.2"
```

main.rs: 
```
use rev_buf_reader;

use rev_buf_reader::RevBufReader;
use std::io::{self, BufRead};

fn main() {
    let data = "This\nis a long\ntest sentence";
    let inner = io::Cursor::new(&data);
    let reader = RevBufReader::new(inner);
    let mut lines = reader.lines();

    assert_eq!(lines.next().unwrap().unwrap(), "test sentence".to_string());
    assert_eq!(lines.next().unwrap().unwrap(), "is a long".to_string());
    assert_eq!(lines.next().unwrap().unwrap(), "This".to_string());
    assert!(lines.next().is_none());
}
```

client/display/follow.rs
```diff
@@ -28,6 +33,26 @@ pub fn follow_local_task_logs(pueue_directory: &Path, task_id: usize, stderr: bo
     // Stdout handler to directly write log file output to io::stdout
     // without having to load anything into memory.
     let mut stdout = io::stdout();
+
+    if let Some(lines) = lines {
```
Nukesor, line 37: A comment explaining what's happening in here would be nice.

lib/src/log.rs
Nukesor: This implementation will be a little less performant as we're no longer operating on the byte level, but also introduce UTF-8 handling. However, performance usually isn't an issue in these cases. A few `μs` on top won't hurt us.

We should also be able to safely use the `utf8_from_lossy` function. It replaces invalid characters with the Placeholder character, which would be done by the terminal lateron anyway.

Nukesor (changes requested):
First of all thanks for your contributions (Including your other PR) :) Welcome to the project.

In general this looks fine, but it would be awesome if you could evaluate whether the rev_buf_reader crate serves this usecase.

This could potentiall us safe a lot of code and replace it with properly tested and well documented code :)

Nukesor:
You can probably ignore most of the comments, if the `rev_buf_reader` works. I only noticed the similarity to the BufReader halfway through the review.

Nukesor:
I went ahead and double-checked the `rev_buf_reader` crate.
In general it looks good to me beside a few little issues.

I created a few PRs (https://github.com/andre-vm/rev_buf_reader/pull/1), let's see how active the maintainer is :)
We could always offer a co-maintainership.

Nukesor:
I just noticed, that this is also missing the logic for streaming the content from the daemon to the client.
It looks like you only implemented the client side local reading.

There's a separate file in `daemon/network/follow.rs`, that's used to stream log output from a remote daemon to clients.

lib/src/log.rs
```diff
@@ -129,22 +130,84 @@ pub fn reset_task_log_directory(path: &Path) -> Result<(), Error> {
...
-///
-/// TODO: This is super imperformant, but works as long as we don't use the last
-/// 1000 lines. It would be cleaner to seek to the beginning of the requested
-/// position and simply stream the content.
 pub fn read_last_lines(file: &mut File, amount: usize) -> String {
-    // Read the lines from the file.
-    // Return a debug string if it fails.
-    let last_lines: Vec<String> = match rev_lines::RevLines::new(BufReader::new(file)) {
-        Ok(lines) => lines.take(amount).collect(),
+    match read_last_lines_as_byte_deque(file, amount) {
```
Nukesor, line 201: After reading through my old docs and the old implementation, I'm not 100% sure if your current implementation changes anything at all.
`rev_lines` also read the file from the back in a BufReader style, but with a small convenience layer on top.

The part about imperformant code mentioned in the docstring isn't about the bufreader implementation, but rather about reading all lines into memory, reversing them and only then streaming them to disk.

This gets troublesome, if users try to read a few thousand lines with possibly gigabytes of data from the logs. This already happened and lead to RAM exhaustion.

A proper performant solution would be to read through the file until X+1 `\n` are found.
The character after the `\n` is then the start of our log output and we could simply pass the handle to a io::copy call with stdout as the sink.
`\n\r` should obviously also be handled

This function would then be changed to something like this:
```
/// Seek to the beginning of a line, which is located `lines` newlines (`\n`) from the end of the file.
/// Jumps to the top of the files, if it doesn't contain enough files.
/// Returns the number of lines that were found.
pub fn seek_to_line_rev(file: &mut File, lines: usize) -> Result<usize> {}
```

drewkett: FWIW, I believe my implementation does in fact read in 4k chunks going backwards. I agree generally it could use more documentation, but I sort of just submitted this as first pass to see if you had interest in using it. That said, I do agree that using that crate probably makes more sense than the implementation I came up with.

Nukesor: It does so, but this is a usual behavior of a BufReader.
The problem isn't the chunked reading, but rather the fact that it collects them into memory before streaming them to stdout.

The performant way would be to seek to the desired position in the file and stream the remaining contents directly to stdout.

It's a "read into memory and copy to stdout" vs. "stream directly to stdout". Reading into memory can be critical with the amount of data that possibly accrues

lib/src/log.rs
drewkett: If I used `u64` though, I believe I would need to do `try_into` or something similar on 32-bit platforms to be safe for the cast to `usize` (obviously the chunk size won't be that big). Thats what I was trying to avoid. Though its possible all my casting is less readable than using `try_into` . So even if we drop 16-bit, to keep my code as is, it could only be `u32`.

lib/src/log.rs
drewkett: Agreed. I'll write one up if we end up keeping this function instead of using the other crate

lib/src/log.rs
drewkett: I see what you're saying now. Yes that is true that I didn't try to address the scenario where a large number of lines are requested. It doesn't strike me as a likely use case, but understood that ideally that would be addressed as well. One thing that stopped me from even looking at that was that I wasn't sure of is whether there is currently a mechanism to stream data between client/server.  Since its message based, I would think there would need to be a mechanism for sending data in chunks from server to client which didn't jump out at me in the code but I also may have just missed it.

That said using that other crate with everything else the same, should at least prevent the extra copy of data in `read_last_lines` since it looked like it stores everything contiguously as opposed to my `VecDeque` based implementation.

drewkett:
Thanks for all the comments @Nukesor. I realize I should have cleaned this up a bit more before just dumping it as a PR. I also  appreciate having you go through my code and point out some of the stylistic things that I do that are not exactly best practices.

lib/src/log.rs
drewkett: Ignore the last comment regarding streaming, of course there is a mechanism to do that. I 

> I went ahead and double-checked the `rev_buf_reader` crate. In general it looks good to me beside a few small issues.
> 
> I created a few PRs ([andre-vm/rev_buf_reader#1](https://github.com/andre-vm/rev_buf_reader/pull/1)), let's wait and see if the maintainer is still active :) We could always offer a co-maintainership.

Just glanced at this PR and it seems like its maintained based on the responses, so I'm going to go ahead and switch to it. I'll do a quick review of that crate as well just to have another set of eyes on.

drewkett: Ignore the last comment regarding streaming, of course there is a mechanism to do that. I 

> I went ahead and double-checked the `rev_buf_reader` crate. In general it looks good to me beside a few small issues.
> 
> I created a few PRs ([andre-vm/rev_buf_reader#1](https://github.com/andre-vm/rev_buf_reader/pull/1)), let's wait and see if the maintainer is still active :) We could always offer a co-maintainership.

Just glanced at this PR and it seems like its maintained based on the responses, so I'm going to go ahead and switch to it. I'll do a quick review of that crate as well just to have another set of eyes on.

drewkett:
In switching to `rev_buf_reader`, how much do you care about performance? The most straightforward implementation as you showed above would allocate a new `String` for every line as it iterates back through the number of lines. An alternative that should be more performant would be to create a function to seek back in the file the requested number of lines and then stream the data as you suggested in another comment. But that implementation would be more complex.

lib/src/log.rs
Nukesor: Ah I see. You're right about that. Casting to usize wouldn't be safe in that case.

Nukesor:
This is quite a fun problem :D

I tinkered around a bit and came up with this solution. I already tried it and it seemed to work like a charm.

```
/// Seek the current file handle to the beginning of the line that's located `amount` newlines
/// from the back of the file.
pub fn seek_to_last_lines(file: &mut File, amount: usize) -> Result<(), Error> {
    let mut reader = RevBufReader::new(file);
    // The position from which the RevBufReader starts reading.
    // The file size might change while we're reading the file. Hence we have to save it now.
    let start_position = reader.get_mut().seek(SeekFrom::Current(0))?;
    let start_position: i64 = start_position.try_into().map_err(|_| {
        Error::Generic("Failed to convert start cursor position to i64".to_string())
    })?;
    println!("Start position: {}", start_position);

    let mut total_read_bytes: i64 = 0;
    let mut found_lines = 0;

    // Read in 4KB chunks until there's either nothing left or we found `amount` newline characters.
    loop {
        let mut buffer = vec![0; 4096];
        let read_bytes = reader.read(&mut buffer)?;

        // Return if there's nothing left to read.
        // We hit the start of the file and read fewer lines then specified.
        if read_bytes == 0 {
            break;
        }

        // Check each byte for a newline.
        // Register all newlines we find.
        // Even though the RevBufReader reads from behind, the bytes are still in the correct
        // order. Since we want to scan from the back, we have to reverse the buffer
        for byte in buffer[0..read_bytes].iter().rev() {
            total_read_bytes += 1;
            println!("char: '{:?}'", *byte as char);
            if *byte != b'\n' {
                continue;
            }

            // We found a newline.
            println!(
                "Found newline at byte: {}",
                start_position - total_read_bytes
            );
            found_lines += 1;

            // We found the requested amount of lines.
            if found_lines == amount + 1 {
                println!("Found {} lines", amount);
                // The RevBufReader most likely already went past this point.
                // That's why we have to set the cursor to the position of the last newline.
                println!("Total read byte: {}", total_read_bytes);

                // Calculate the distance from the start to the desired location.
                let distance_to_start = start_position - total_read_bytes + 1;
                // Cast it to u64. If it somehow became negative, just seek to the start of the
                // file.
                let distance_to_start: u64 = distance_to_start.try_into().unwrap_or(0);
                println!("Distance: {}", distance_to_start);

                // We can safely unwrap `start_position`, as we previously casted it from an u64.
                if distance_to_start < start_position.try_into().unwrap() {
                    // Seek to the position.
                    let inner = reader.into_inner();
                    let current_position = inner.seek(SeekFrom::Start(distance_to_start))?;
                    println!("Final position: {}", current_position);
                }

                return Ok(());
            }
        }
    }

    Ok(())
}
```

Nukesor:
I went ahead and implemented the new seeking logic in a separate PR.
This PR also fixes some other log related issues and adds proper tests for this behavior.
https://github.com/Nukesor/pueue/pull/278

Adding your feature with these changes should now be trivial :)

drewkett:
Definitely makes this change simpler 👍. I cleaned up the commit history and stripped it down to just this change. Let me know if I missed anything

Nukesor:
Perfect, this looks good to me :)

Thanks for your contribution and sorry for the constant back and forth ;D

Nukesor (approved):

PR #285, opened 2022-01-15
A few clippy fixes on Windows

drewkett:
This just fixes up clippy warnings on Windows that I noticed when working on another PR. They don't seem essential, but useful to just clear the warnings.


Nukesor (commented):
Nice, Thanks :)

Issue #239, opened 2021-09-12
Filter log output by stdout/stderr, and time


Nukesor:
The PR https://github.com/Nukesor/pueue/pull/286 combined the two streams of a task's output back into a single output.

It would now be awesome to get the possibility to filter for either `stderr`/`stdout` and maybe even time.

For this to work, we would need some way to combine the two output streams without loosing temporal/chronological information. Furthermore the resulting output should be some kind of dataformat that should be structured and supports streaming.

The requirements to this feature are:

1. Task log output should be in the correct chronological order.
2. Log output should be error-resistant (partial writes shouldn't break the whole log)
3. We have to be able to differentiate between `stderr` and `stdout` output. I.e. we have to be able to sort by it.
4. Remote streaming should still be performant (dataformat should support streaming)
5. The dataformat should have support for meta-data such as time-stamps

We should prevent to load whole files into memory. If we have to load them into memory, we should work with streams and utilize in-memory compression wherever possible.
Log files can be huge and we already got reports of crashing PCs due to high RAM usage.

Help, suggestions or even PRs are very welcome!

Originating issue: #228

LovecraftianHorror:
> Help, suggestions or even PRs are very welcome!

I would be open to contributing. It will take me a little bit to get more familiar with the codebase, but I can send an update in a couple of days if you're fine with that

Nukesor:
Feel free to tinker around a bit :)

I would love to your approach on this problem.

Nukesor:
Once you have some visdion on how to implement this, it would be awesome to hear about it.

Especially how you plan to combine the two streams into a singular stream whilst loosing as little (or no) temporal information as possible. And what kind of dataformat you have in mind, since it has to support streaming.

I stumbled upon [os_pipe](https://github.com/oconnor663/os_pipe.rs) when researching on how to merge pipes.
Not sure how useful this might be, though.

I mind such an implementation would probably need a thread per pipe per task?
An alternative would probably be some `async` impl on top, so we don't spawn a huge amount of actual threads but rather let the tokio scheduler handle the rest.

The model of one thread/async-task per pipe might get problematic though, if a user runs a hundred tasks in parallel and each task produces a huge amount of logs. I.e. what happens if the log threads/async-tasks don't manage to keep up with the task's output. 
It would also be interesting to investigate how this behaves if the system has a high load in general.

The best solution would obviously be to have a solution which doesn't need threading, if something like this exists.

LovecraftianHorror:
[...]

PR #286, opened 2022-01-17
Use a single file for task output

Nukesor:
This PR is a prototype for a pretty large change of Pueue's internal working.

I've been thinking a lot about how Pueue's log handling should work in the first place.
Originally, I thought that it would be super handy to be able to filter log output by stdout/stderr.
However, by splitting stderr and stdout one loses all chronological context between those two.

This PR now merges both streams back together into a single file.
This makes using `log` and `follow` much easier to use and logs are now in correct chronological order, but one can no longer filter by `stdout` or `stderr`.

Since this project became quite big, I would like to get your opinion on this change.
What do you think about it? Just leave a thumbs-up if you're happy with this change and please leave a comment if you have any additions/doubts/thoughts about this.

Sorry for pinging all of you, but I would really like to get some feedback on this, since this is a huge breaking change.
@taylor1791 @KaindlJulian @Lej77 @sourcefrog @tinou98 @quebin31 @JP-Ellis @soruh @twillingham @Sh3Rm4n @dovahcrow @LovecraftianHorror @dadav @Mephistophiles @printfn @Cycatz 

I would love to get a single stream while still keeping the possibility to filter by stderr/stdout and maybe even time, but this turns out to be quite complicated and I won't be able to finish such a feature in the near future.
The tracking issue is #239.

As a result, I thought that it might be a good compromise to just merge the streams for now.
v2.0 has been in development since a few months now and a huge amount of changes has accrued.
The next implementation might then be added in a future v3.0 release.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Allow to set the amount of parallel tasks at group creation by [Spyros Roum](https://github.com/SpyrosRoum) [#245](https://github.com/Nukesor/pueue/issues/249).
 - When calling `pueue` without a subcommand, the `status` command will be called by default [#247](https://github.com/Nukesor/pueue/issues/247).
 - Add the `--group` parameter to the `pueue clean` command [#248](https://github.com/Nukesor/pueue/issues/248).
-- Add `stdout_path` and `stderr_path` as template parameters for callbacks [#269](https://github.com/Nukesor/issues/269).
+- Add `output` for a task's log output as template parameters for callbacks [#269](https://github.com/Nukesor/issues/269).
 - Add `--lines` parameter to `pueue follow` to only show specified number of lines from stdout before following [#270](https://github.com/Nukesor/pueue/issues/270).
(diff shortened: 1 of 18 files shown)


sourcefrog:
I haven't read the diff yet, but from the description this sounds good.

I think for me the most common case of caring to split stdout from stderr is when I direct stdout into a file while looking at errors on stderr. In that case I think I could and probably would just redirect it within the command given to pueue.

I'll come back and look at the diff.

JP-Ellis:
Thanks so much for all the work you're putting into this!

I am very much for the merging of the two streams.  Since pueue's purpuse is to run tasks asynchronously, the output of programs should generally only be informational / logs.

While I don't use it in this way, I could imagine that this causing an issue for people who produce output files from stdout.  For example, if a program produced a json file in stdout and printed warnings in stderr, the merging of the two streams would produce an invalid json file; however as @sourcefrog mentioned this can be easily fixed by redirecting the output within the command given to pueue.

Nukesor:
Well, I guess this can then be merged soon :) 

@sourcefrog I like the idea of users still being able to split stdout and stderr themselves by simply using the shell's tools. I've never thought about this, but theoretically users could already combine the streams previously via `2>&1` in the past.

PR #297, opened 2022-03-04
Don't inherit environment variables from daemon process

drewkett:
Currently, the environment variables associated with a task are added to those of the daemon process on launch. Use `env_clear()` to start with a clean environment before adding the task environment variables

I wasn't sure if this should merge to `main` or `development`

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- Commands no longer inherit environment variables from the daemon process
+
 ## [2.0.0] - 2022-02-18
 
 This release marks the second stable release of Pueue.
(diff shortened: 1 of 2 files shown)


Nukesor:
I would consider this a bug, so a merge against `main` is correct :)

Nukesor:
Good catch, I didn't expect the environment variables to be automatically inherited, but I guess it makes sense that this is the default behavior.

Thanks for discovering and fixing this :)

I want to write an integration test before merging this. Once I've done this and it fails as expected on the current main, I'll merge your PR :D. You don't have to do anything on your own, but this definitely is something that should be properly tested.

drewkett:
Related to this, I had the thought was whether the environment variables should also be set for callbacks. I didn't actually test them but the code seemed to indicate that they are not set.

Nukesor:
Nope they're not set yet.

What's a good use-case for this?

drewkett:
Honestly I don't have one. I ran into this original issue cause I had set an environment variable in the terminal that I then used to launch pueued. And that caused an error in a task since it tried to use that environment variable but it was incorrect. So my thinking with the callback is that the more correct environment for the callback is probably the environment of the task versus whatever the environment is for the daemon.

Nukesor:
Perfect :) 

I think the daemon environment is good enough for callbacks for now. Let's keep it that way until somebody runs into some problems and comes up with a proper use-case for using a different environment.

Until then, thanks for your fix! I managed to write a proper integration test for this bug and your change indeed fixes the problem at hand.

PR #310, opened 2022-05-25
add gnu parallel

chapmanjacobd:
I've created something [very similar to this in the past](https://github.com/chapmanjacobd/que) before I was aware of GNU Parallel. While GNU Parallel does have a learning curve, if there is a simple example then people might see that it could work for them

## Checklist

- [n/a] I included a new entry to the `CHANGELOG.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,29 @@ There seems to be the need for some solution that satisfies all these points men
 
 ## Similar Projects
 
+**GNU Parallel**
+
+A robust and featureful parrallel processor with text-based joblog and n-retries (but it also integrates with SSH and databases if you have the need to farm out jobs to multiple machines). [Link to project](https://www.gnu.org/software/parallel/parallel_tutorial.html)
+
+Sequential jobs example:
+
+```sh
+tee example1.sh 
+#!/bin/bash
+printf 'cmd-to-run "%s"\n' "$@" | parallel -j1 --joblog p.log --resume --retries 3
+```
+
+Running this `example1.sh taipeifilmcommission vihartvihart socallinuxexpo` would start the following jobs and re-run them up to 3 times if they failed:
+
+```
+cmd-to-run taipeifilmcommission
+cmd-to-run vihartvihart
+cmd-to-run socallinuxexpo
+```
+
+- If you want parallel change j1 to j3 or remove it altogether and GNU Parallel will chose for you using the number of CPU cores you have
+- You can also add the -u flag (`parallel -u`) if you prefer to see all the output as the commands run (instead of printed when each command finishes)
+
 **nq**
 
 A very lightweight job queue systems which require no setup, maintenance, supervision, or any long-running processes. \
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ There seems to be the need for some solution that satisfies all these points men
 
 **GNU Parallel**
 
-A robust and featureful parrallel processor with text-based joblog and n-retries (but it also integrates with SSH and databases if you have the need to farm out jobs to multiple machines). [Link to project](https://www.gnu.org/software/parallel/parallel_tutorial.html)
+A robust and featureful parrallel processor with text-based joblog and n-retries. [GNU Parallel](https://www.gnu.org/software/parallel/parallel_tutorial.html) is able to scale to multi-host parallelization and has complex code to have deep integration across different tools and shells, as well as other advanced features. `Pueue` differentiates itself from GNU Parallel by focusing more on visibility across many different long running commands, and creating a central location for commands to be stored, rather than GNU Parallel's focus on chunking a specific task.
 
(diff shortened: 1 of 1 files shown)


chapmanjacobd:
Fair enough. I feel like GNU Parallel's documentation is too academic and I doubt they would be willing to make it more user-friendly. I'll move my code example into a gist and make the PR here more generic and formal

chapmanjacobd:
I'm not super familiar with `pueue`. It would be helpful for me and others if you could expand on how `pueue` differentiates itself from GNU Parallel--especially your perspective on GNU Parallel's limitations and drawbacks which `pueue` is able to solve. Feel free to address that here or by creating a commit in this PR's branch. If you only comment here I'll add some of your thoughts to the PR if it's okay

Nukesor:
Scope of the project:

gnu-parallel is mostly designed as a heavy duty parallel executor.
Pueue however aims to be a minimalistic scheduler.
This means that there're a many more features in Pueue and also many more in gnu-parallel.
But most importantly, the focus of each program is a completely different one.

Many missing features may be possible to achieve with gnu-parallel and additional tools, but this is just ductape.

The idea is to create a nice and convenient CLI tool for managing shell tasks.
This means a few things:

1. The cli should be intuitive!
2. The output should be pretty (colored) and well-organized under all circumstances.
3. It should also be easily scriptable (machine readable output).


Related links: https://github.com/Nukesor/pueue/wiki/FAQ

# Pueue features/defaults

- State management via global daemon.
    * Queue(s) and tasks are global.
    * Recoverability and no data loss on crash/shutdown
    * Gnu-parallel doesn't manage a global state. Restart your PC and your queue is gone.
    * As far as I can see, the tasks are also bound to the terminal?
        This means, your commands are gone as soon as you close your terminal, unless you use the tmux feature, which spawns a new tmux shell for each command?
- Command output is managed in files by default.
    * No log loss on crash/shutdown.
    * Easy inspection from anywhere on the same machine.
    * Gnu-parallel caches cmd output in randomly named files in `/tmp/` by default, resulting in data loss on crash.
- Easy editing + rescheduling of tasks' commands/paths.
- Groups! Multiple queues can run in parallel.
    * Most commands (pause/start/kill) can be run on a whole group.
    * Each group has their own count of parallel tasks.
    * Provides simple ENV variables to allow mapping of pueue workers/groups to external resources.
        E.g. Singular GPUs in a cluster of worker machines.
- Scheduling of messages at a specific time (with human time input parser).
- Dependency graphs for tasks.
- **Easy** control over running commands (pause/restart/kill).
- **Easy** inspection and manipulation of the queue (e.g. stash tasks for later usage).
- `Json` output for `status` and `log`
- `cbor` API + protocol, which can be used by other programs (though quite low-level).
- Callback hook on finished tasks (notifications, external tool integration, etc.).

# Gnu-parallel-only features

Only to name a few:

- Automatic creation of multiple tasks at once via special syntax (their syntax is crazy powerful).
- Multi host parallelization. Pueue is designed to be single-host.
- Tight integration of external tools (e.g. tmux, sql)
    It's specifically not one of Pueue's design goals to build integration for external tools.
    This significantly increases the project's complexity and maintenance burden.
- Integration with `nice`.    
- parameter level or file-block level parallelization.
    Gnu-parallel offers a lot of possibilities to chunk some kind of workload into many parallel tasks.
- Lots of stuff. I don't want to read the whole Gnu documentation, just to enumerate all possible features.

Gnu-parallel is definitly a nice tool to have, but it's most certainly build for another usecase.
It's purpose is to easily split any task (or group of tasks) into highly parallelizable chunks of work.

Pueue on the other hand is designed to handle various, often completely different, long running commands.
By long running commands, I mean queues that span across several weeks (e.g. hundreds of reencoding tasks).

These are two completely different usecases. \
It's much more important (at least for me) to have an easy-to-use specialized tool for each usecase.
In my case, this usecase would be, managing lots of long-running processes on my server and desktop.



If you like, you can incorporate this post in the project.
However, I think that the wiki is better suited for this format.
The README already has a lot of info in it.

chapmanjacobd:
Thank you @Nukesor that is very helpful. I updated the PR to reflect some of that insight

But to clarify some of your questions maybe I can help:

> Restart your PC and your queue is gone.

This is true if you don't use the `joblog` flag. The joblog keeps track of which jobs have completed but you are correct you would also need to create a file (one line per command) to manage the continuation of tasks across shells or across reboot.

> spawns a new tmux shell for each command

I'm not sure but I believe GNU Parallel somehow uses the same shell/environment--you are right there is a lot of complex code to handle that

> Gnu-parallel caches cmd output in randomly named files in /tmp/ by default, resulting in data loss on crash.

very good point

Nukesor:
MR looks good to me :) 

Thanks for the contribution!

PR #313, opened 2022-06-04
Fix config dir and PID path

aeveris:
This MR removes the `.yml` suffix from the default configuration directory, reverting it to the expected `~/.config/pueue/` and also uses the proper path for PID file removal during shutdown. 

## Checklist

- [X] I included a new entry to the `CHANGELOG.md`.
- [X] I checked `cargo clippy` and `cargo fmt`. The CI will fail otherwise anyway.


Nukesor:
Damn. This just slipped through.
Sadly this isn't something that's easy to test either..

Thanks! I'll release a patch momentarily

PR #319, opened 2022-07-04
Use style_text to style group header

mjpieters:
This ensures that the bold styling is only applied when on a TTY.

## Checklist


Nukesor:
Thanks! This one slipped through.

Issue #315, opened 2022-06-22
[BUG] Window process test is flaky


Nukesor:
The Windows CI test suite flakes from time to time.

It's always the same error. Sadly I don't know much about windows process handling.
Any help is much appreciated.

Here's the Error:

```
running 17 tests
test network::message_handler::clean::tests::clean_normal ... ok
test network::message_handler::clean::tests::clean_normal_for_all_results ... ok
test network::message_handler::clean::tests::clean_only_in_selected_group ... ok
test network::message_handler::clean::tests::clean_only_successful_only_in_selected_group ... ok
test network::message_handler::remove::tests::normal_remove ... ok
test network::message_handler::clean::tests::clean_successful_only ... ok
test network::message_handler::switch::tests::switch_double_dependency ... ok
test network::message_handler::switch::tests::switch_invalid ... ok
test network::message_handler::switch::tests::switch_normal ... ok
test network::message_handler::switch::tests::switch_task_with_dependant ... ok
test network::message_handler::switch::tests::switch_task_with_itself ... ok
test network::message_handler::remove::tests::removal_of_dependencies ... ok
test platform::windows::process_helper::test::test_normal_command_children_are_killed ... ok
test platform::windows::process_helper::test::test_normal_command_is_killed ... ok
test platform::windows::process_helper::test::test_shell_command_is_killed ... ok
test platform::windows::process_helper::test::test_shell_command_children_are_killed ... FAILED
test platform::windows::process_helper::test::test_spawn_command ... ok
failures:
---- platform::windows::process_helper::test::test_shell_command_children_are_killed stdout ----
Error: 2 processes were expected. Last process count was 1
thread 'platform::windows::process_helper::test::test_shell_command_children_are_killed' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\library\test\src\lib.rs:186:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
failures:
    platform::windows::process_helper::test::test_shell_command_children_are_killed
test result: FAILED. 16 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 52.23s
error: test failed, to rerun pass '-p pueue --lib'
Error: The process 'C:\Users\runneradmin\.cargo\bin\cargo.exe' failed with exit code 101
```

[...]

PR #320, opened 2022-07-04
Add --color cli switch

### LINKED ISSUE #318: Add --color switch to control use of colours

Currently, pueue uses `stdout().istty()` to determine when to use colours in the output. However, there are use-cases for outputting colour even if not connected directly to a TTY; e.g. when using `less` with the `-R / --RAW-CONTROL-CHARS` switch, or `watch` with the `-c/--color` switch, which usually then pass on the output to a TTY.

Many CLI programs let you control when they use colours with a `--color` switch; e.g. `ls` has `--color=never`, `--color=always` and `--color=auto`; the current behaviour of `pueue` is akin to `--color=auto` (with a few exceptions that are always using colours, probably a bug).

Please add such a switch to pueue too!

Not using `watch` and baking up my own `while :; do pueue status; sleep 2; printf "\033[2J\033[1;1H"; done` loop, which has an annoying flicker as the screen is redrawn entirely (`watch --color pueue` is much, much smoother).

mjpieters:
This adds a --color switch to the `pueue` command, controlling when output will
include ANSI styles (colors, bolding). The default, `auto` matches the current
behaviour: style when connected to a TTY.

This PR refactors the way styling is handled (consolidating color choices and
styling into a single location struct), and this struct then tracks if styling
can be output or not.

This would fix #318.

## Checklist


mjpieters:
I wonder why the build failed, but mostly why the target is `aarch64-apple-ios`. I'm working on a M1Pro, target `aarch64-apple-darwin`, perhaps that was intended?

mjpieters:
> I wonder why the build failed, but mostly why the target is `aarch64-apple-ios`. I'm working on a M1Pro, target `aarch64-apple-darwin`, perhaps that was intended?

Ah, no, it's more of a general issue with the rust toolchain, as the same issues appeared on the main branch.

Nukesor:
Thanks for your contribution!

Yes, the CI seems to be broken for mac due to some changes in Rust v1.62.
I'll have a detailed look at your PR as soon as I find some time. I already took a first glance and the changes looked quite intuitive :)

mjpieters:
> I'll go over it once more as I see that some of the client/display/state code uses comfy-table directly, which means that my changes to how the dark-theme colours are used may be affected.

I did indeed break dark styles for tables; I see two options to remedy this:

- Revert the part of my refactor where the OutputStyle struct takes over swapping colours, putting the responsibility back on any code using colours.
- Add a `style_cell()` method to the struct that lets you create styled comfy-table cells.

I'd pick the latter myself, so I'll just update the middle commit to do that. The other option would be to drop that commit.

mjpieters:
Is the Windows test failure a flakey test? It also failed for a [build of my toolchain PR](https://github.com/Nukesor/pueue/actions/runs/2624304150) but not when that PR was merged, but there the windows build was cancelled due to a codecov upload failure.

Nukesor:
@mjpieters Yep, the windows test is a flaky one :D There's already an open ticket on this:

https://github.com/Nukesor/pueue/issues/315

I tried to increase the wait threshold several times, but this seems a different problem at hand. The original implementor of the windows code didn't know about it either and just suggested to increase the wait threshold.

client/display/style.rs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        let mut styled = style(text);
+        if let Some(color) = color {
+            styled = styled.with(match color {
+                Color::Green => self.green,
+                Color::Red => self.red,
+                Color::Yellow => self.yellow,
+                _ => color,
+            });
```
Nukesor, line 62: You could extract this duplicate match block into a separate `self.color` or `self.shade` function.

client/display/state.rs
```diff
@@ -207,29 +210,25 @@ fn print_table(tasks: &[Task], colors: &Colors, settings: &Settings) {
...
         }
-        row.add_cell(Cell::new(&task.path.to_string_lossy()));
+        row.add_cell(style.styled_cell(&task.path.to_string_lossy(), None, None));
 
         // Add start and end info
         let (start, end) = formatted_start_end(task, settings);
-        row.add_cell(Cell::new(start));
-        row.add_cell(Cell::new(end));
+        row.add_cell(style.styled_cell(start, None, None));
+        row.add_cell(style.styled_cell(end, None, None));
```
Nukesor, line 232: I would rather like to not call `style.styled_cell` on Cells that don't need any styling.

Actually, most of `style.styled_cell` calls in here can be replaced with simple `Cell::new()` calls.
There're also a few in the `log.rs` file.

Nukesor (changes requested):

Nukesor:
Once more, thanks a lot for your contributions! I'm always very happy when I get such high-quality contributions!

I'm also impressed by your detailed review and improvement of this project's CI! I definitely learned a few new things :)

Much appreciated!

Nukesor:
FYI I decided to ignore the flaky windows test for now.

Previously it only occurred every now and then, but recently it showed up much more consistently. My initial assuption would be that this has something to do with the current load on the Github's windows CI machines.

client/client.rs
```diff
@@ -110,7 +115,7 @@ impl Client {
 
         Ok(Client {
             settings,
-            colors,
+            style: colors,
```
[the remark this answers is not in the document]
mjpieters: Good catch, this was an error (I had renamed it in the refactor commit, but lost the rename)

client/display/state.rs
mjpieters: Sure, I can revert that. I had done this because it may be a good idea to consistently _not_ use Cell::new() so you don't, in future, forget to replace these with `style.styled_cell()` calls when you do want to add formatting to those cells.

client/display/style.rs
mjpieters: Excelent idea!

client/display/state.rs
Nukesor: I agree that this might prevent somebody forgetting `style.styled_cell()` at some point, this is something I'll would to take care of when reviewing.

In this case it feels like the benefits of better readability/less obfuscated code surpass the benefit of consistency.
Of course, this is a very subjective opinion based on personal code style ;D

mjpieters (commented):
This is _your_ project, so I'll happily follow your style here.

And I agree that the `Cell:new()` API is cleaner. I didn't want to completely replicate that API (return a custom `Cell` struct which ignores chained `fg()` and `add_attribute()` calls when colours are disabled), perhaps this is something that the upstream `comfy_table` project should provide.

Nukesor (approved):

Nukesor:
Review adjustments look good to me :)
I also tested the changes manually and everything worked just as expected.

Thanks again for your contribution!

PR #322, opened 2022-07-06
CI: Add a cache for cargo to speed up actions

mjpieters:
The cache step here caches the cargo output as per the [recommended
settings](https://github.com/actions/cache/blob/main/examples.md#rust---cargo),
but with the cross-compilation target as part of the cache key, and a restore key
to take advantage of a partial cache should the cargo lock change only marginally.

## Checklist

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -23,6 +23,7 @@ jobs:
     name: Test on ${{ matrix.os }} for ${{ matrix.target }}
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         target:
             - x86_64-unknown-linux-gnu
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,6 +23,7 @@ jobs:
     name: Test on ${{ matrix.os }} for ${{ matrix.target }}
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         target:
             - x86_64-unknown-linux-gnu
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -74,6 +74,18 @@ jobs:
           components: llvm-tools-preview
           target: ${{ matrix.target }}
 
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-${{ matrix.target }}-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-${{ matrix.target }}-
+
       - name: cargo build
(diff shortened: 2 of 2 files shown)


mjpieters:
And because I rebased this PR onto main after @Nukesor merged my other CI PR, we get a neat demo of how effective the cache is :-)

The initial test workflow run for this PR, the `cargo build` steps [took up to 5 minutes](https://github.com/Nukesor/pueue/actions/runs/2624499686), and after a force push the same workflow [took 20 seconds in the best case, 1 min 40 in the worst](https://github.com/Nukesor/pueue/actions/runs/2624557278).

Nukesor:
Haha :D

You're having a field day over here! Thanks for bringing the CI up to speed :)  
Caching looks good. A good 50% time reduction between runs seems reasonable.

Regarding `fail-fast`, there's no specific reason why this isn't used yet.
To be honest, I remember looking this up when migrating from travis to github actions, but IIRC they didn't have this feature back than :D Maybe it got added in the meantime,

Feel free to add it to all action definitions! That would definitely be quite handy.

mjpieters:
To be clear: you have fail-fast _now_, because that's the default. It's why in [this build](https://github.com/Nukesor/pueue/actions/runs/2623398409), all the steps except *Test on macos-latest for aarch64-apple-ios* were cancelled; the ios build failed, and that triggered a fast fail path that cancelled the other matrix members.

Setting `fail-fast` to `false` would mean those other builds are permitted to continue. Like when the coverage upload for one of the builds fails (not always a fatal issue!), you'd still see if the other builds at least can complete.

I'll add the switch to this PR.

mjpieters:
Also: I don't think the linter steps need to be run across all targets. cargo fmt and cargo clippy output won't vary, will they? Or are parts of the codebase conditionally ignored based on platform?

Nukesor:
> Also: I don't think the linter steps need to be run across all targets. cargo fmt and cargo clippy output won't vary, will they? Or are parts of the codebase conditionally ignored based on platform?

That's the case. At least clippy only looks at the pieces of code/features that're currently included and actually compiled.

I would really like to keep this in the CI, as I otherwise won't notice any linting issues in PRs, as I'm always on Linux.

mjpieters:
> That's the case. At least clippy only looks at the pieces of code/features that're currently included and actually compiled.

Right you are! Then the matrix there is necessary and helpful.

> For now, clippy and fmt isn't actually executed on all platforms, but it'll be in the future, as soon as all features are implemented for the apple platforms.

For what it's worth: `cargo fmt --all -- --check` and `cargo clippy --tests --workspace -- -D warnings` pass on my Macbook Pro (`target=`aarch64-apple-darwin`).

I haven't quite managed to cross-compile to `aarch64-apple-ios` or `x86_64-apple-darwin` yet, so I can't confirm if this applies to those as well. That's because the `ring` crate needs to [cross-compile C++ code too](https://github.com/briansmith/ring/blob/main/BUILDING.md#cross-compiling), and that's not something I've done before. Not as big a challenge as cross-compilation from Mac to Linux, though.

However, I don't see why the codebase wouldn't pass on those platforms right now.

PR #323, opened 2022-07-06
CI: Lint on all platforms and targets

mjpieters:
Running clippy and fmt on all targets and platforms should be fine.

## Checklist

- [X] I picked the correct source and target branch.
- [ ] I included a new entry to the `CHANGELOG.md`.
- [ ] I checked `cargo clippy` and `cargo fmt`. The CI will fail otherwise anyway.
- [ ] (If applicable) I added tests for this feature or adjusted existing tests.
- [ ] (If applicable) I checked if anything in the wiki needs to be changed.

PR #324, opened 2022-07-06
CI: Lint on all platforms and targets

mjpieters:
Running clippy and fmt on all targets and platforms is fine, bar one issue with
Windows that I fixed.

Since clippy needs to compile a bunch, I've added a linting-specific cache to
speed up this step.

(this PR was recreated from #323, now from an actual branch in my fork. Mea
Culpa).

## Checklist

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -19,9 +19,10 @@ on:
 
 
 jobs:
-  publish:
+  linting:
     name: Lint on ${{ matrix.os }} for ${{ matrix.target }}
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


mjpieters:
TBH I am not sure why clippy complains that `helper::get_shared_settings()` is now unused. It's too late at night to go hunt that down now, so I'll look again tomorrow.

Nukesor:
I managed to find the issue :)

The function was used in the unix socket test file, which only executed the test on unix.
As a result, the helper module was declared but not used.

https://github.com/Nukesor/pueue/pull/328

mjpieters:
> I managed to find the issue :)
> 
> The function was used in the unix socket test file, which only executed the test on unix. As a result, the helper module was declared but not used.

That's the thing that confused me; it was not only referenced in the unix socket test, but also in `lib/tests/tls_socket.rs::test_tls_socket`, but clearly clippy didn't see it as in use there. And I could not for the life of me figure out why that test wouldn't run on Windows. Oh well.

Nukesor:
Yep. Integration tests are compiled as a stand-alone binary for each top-level file in the `tests` directory`.

-> Lints might be executed multiple times on the same file from a different perspective. It's a bit confusing :D

PR #357, opened 2022-09-07
Show task label in log output

mjpieters:
If a task has a label, include it in the log information header.

## Checklist


mjpieters:
This PR addresses #357

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [2.1.1] - unreleased
 
+- pueue log output now includes the task label, if any. [#355](https://github.com/Nukesor/pueue/issues/355)
+
```
Nukesor, line 10: Could you move this to the `### Added` section

tests/client/log.rs
```diff
@@ -70,6 +70,24 @@ async fn read_truncated_log(#[case] read_local_logs: bool) -> Result<()> {
     Ok(())
 }
 
+/// If a task has a label, it is included in the log output
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn task_with_label() -> Result<()> {
+    let daemon = daemon().await?;
+    let shared = &daemon.settings.shared;
```
Nukesor, line 77: Nice!

What was your experience with the client tests so far?
I tried to work with existing snapshot comparison libraries, but it turned out that they didn't work for this test setup.

This is why Pueue is using a custom built templated snapshot comparison (everything around `assert_stdout_matches`).

Was it intuitive to write this test or did you run into any obstacles?
In case it's unintuitive, I'll try to add a few docs.

Nukesor (changes requested):
One question and one nitpick, otherwise this looks good to me :)

CHANGELOG.md
Nukesor: There's so much stuff in the current release, you can change the next release from `2.1.1` to `2.2.0` as well while you're at it :D 

This somehow mutated from a point release to a proper minor feature release.

tests/client/log.rs
mjpieters: It was a little unintuitive because whitespace mattered more than I expected. That was harder to debug because there is no diff view of what was missing (spaces after the label portion). Otherwise, it was simple enough to discover because I was looking for tests in the same area.

I don't think the whitespace parts matter, really, but it's up to you if whitespace at the ends of lines should be stripped or not before comparing. To make it easier to figure out why a snapshot doesn't match, I'd look at [`similar-asserts`](https://docs.rs/similar-asserts/), perhaps?

What issues did you have with existing snapshot libraries?

Nukesor: > What issues did you have with existing snapshot libraries?

IIRC the main problem was missing templating support for the snapshots. I tried to work around that, but there're just too many things that depend on the temporary directory or time.
Since we spawn actual processes, it's also not easy to mock system-specific APIs, such as time.

Nukesor: You're right, I also struggled with the whitespaces while writing the tests.
I kept them to ensure that the output really is identical to the output provided by Pueue, but I guess this could be a case where developer experience should be preferred to correctness.

I might go ahead and strip the whitespaces.

Nukesor: `similar-asserts` looks great!
I just added it to the snapshot error messages and it makes catching errors much easiert:

https://github.com/Nukesor/pueue/commit/721eef646f3b42601115a7ecbcdf800b9dfb955c

```
![image](https://user-images.githubusercontent.com/3322822/189139199-58984b33-64df-4bbc-85ee-94aac97cc320.png)

Nukesor (approved):

Nukesor:
LGTM :)

Thanks for your contribution!

PR #361, opened 2022-09-10
change: Migrate from serde_cbor to ciborium

### LINKED ISSUE #360: [Debt] Move away from the `serde_cbor` crate
The `cbor` crate has been officially archived.

The owner suggests to move to either of these alternatives:
- https://crates.io/crates/ciborium
- https://crates.io/crates/minicbor

# Ciborium

[According to the author](https://github.com/enarx/ciborium#compatibility-with-other-implementations), this one is compatible with `serde_cbor` and could thereby be used as a drop-in replacement.

# Minicbor

This one is, as the name suggests, rather minimalistic. It might be enough for our use case, but we don't really aim for no-std environments and we also don't have to be minimalistic.

Nukesor:
Resolves #360 

This commit moves all logic from `serde_cbor` to `ciborium`.
The test suite seems to pass and ciborium is also explicitly advertised as being compatible with serde_cbor.

I would like to do some further analysis and proper testing though.
- [ ] Run the current daemon against a client with the `cbor` implementation


Nukesor:
Closed, as ciborium seems to be abandoned

PR #374, opened 2022-10-13
Give each task a separate process group

### LINKED ISSUE #372: FR: give child processes a group ID, then use that to kill the process tree

Killing child processes of a spawned task is still problematic, e.g. when using a bash script that starts other commands that start their own, etc. (children of children of children). `pueue kill -c <task_id>` won't kill the whole tree, because the parent id for children's children will be that of the intermediates.

Start tasks with a [new process group id](https://en.wikipedia.org/wiki/Process_group), then use the progress group id to kill the tree. It's the tactic that [the `watchexec` project](https://watchexec.github.io/) uses, and they have extracted the functionality needed to do this in Rust as the [`command-group` library](https://docs.rs/crate/command-group/latest).

To quote the [project blurb](https://watchexec.github.io/#command-group):

> Run commands in process groups. Stop creating zombies. Avoid unwanted detachment. **Regain control!**

If tasks were executed using `.group_spawn()` (provided by `command-group`) instead of the stdlib `.spawn()`, `child.kill()` would kill the whole group, not just the top-level process.

Internally, on POSIX systems, the child process calls [`setsid()`](https://docs.rs/nix/latest/nix/unistd/fn.setsid.html) immediately after forking, and `GroupChild.kill()` then uses [`pgkill()`](https://docs.rs/nix/latest/nix/sys/signal/fn.killpg.html) to send the SIGTERM signal to all processes in the process group.

I actually tried to kill a process tree using the group id, which killed the pueue daemon too; it's all the same group. :-(

I've considered adding a [suitable bash signal handler](https://stackoverflow.com/a/2173421) to my scripts, but that's tricky to get right and would need to be added to *all* my scripts, plus I need to [run my scripts with `set -m`](https://stackoverflow.com/a/45112755) to ensure that they have a separate group id from pueue. Or I'd need to spawn all my tasks with `setsid` as the parent task, then manually kill the child processes. Ugly all round.

#### Additional context

The library supports POSIX process groups and winapi job objects, covering Linux, MacOS and Windows. :-)

If a process in the task process tree were to change group ids (using `setsid` or `setpgid`), all bets are off, but that's already the case.

#### Process groups are how shells manage jobs

Process groups are how a [_shell_ handles job control](https://en.wikipedia.org/wiki/Job_control_(Unix)#Implementation), when running interactively.  _In normal shell operations_, it is the use of PGs that make it possible to exit a hierarchy of processes. To keep terminology simple, lets call that the 'interactive shell', and any non-interactive shell process a 'shell script'.

So, in an interactive shell, when you start a command, be that a shell script or something else, the interactive shell creates a new PG for that command. From there on out, it controls that PG using signals, e.g. when you put the command in the background (with `Ctrl-Z` or by starting it in the background with `&`), killing that command is done by sending `SIGKILL` to the whole PG. If the interactive shell exits (say, because something send it a SIGKILL signal), it'll signal each PG it started with `SIGHUP`, wait for those PGs to exit, before exiting itself.

It is only when you _don't_ run a shell interactively that this doesn't happen. That's because that would interfere with the normal operation of a parent (interactive) shell, you don't want the process tree started by a shell script to start a new PG as that would remove control from the interactive shell.

Note: It is actually _surprisingly hard_ to handle signals in a bash script, because _unless you put commands it starts in the background_, a shell script [**won't handle signal hooks set with `trap` until any active child process exits**](https://tldp.org/LDP/Bash-Beginners-Guide/html/sect_12_02.html#sect_12_02_02). You _can_ enable 'monitor mode' in a shell script with `set -m`, in which mode _background_ processes started by the shell script will be given a new PG, but this doesn't happen for foreground processes. This is why you can't just kill the parent shell script and hope that the child processes come along. In fact, monitoring mode (`set -m`) is mostly used to [prevent the interactive shell from killing a child background process when the interactive shell exits](https://unix.stackexchange.com/questions/196603/can-someone-explain-in-detail-what-set-m-does).

Moving to PGs means the `--child` switch can be deprecated, as it should be the _default_ that all signals sent by pueue to a task go to the whole PG, just like [standard interactive shell job control](https://unix.stackexchange.com/questions/594141/what-determines-whether-a-scripts-background-processes-get-a-terminals-sigint), where `kill %jobid` will [send the signal to the PG](https://github.com/bminor/bash/blob/f3a35a2d601a55f337f8ca02a541f8c033682247/builtins/kill.def#L232-L248).

Given that _the default behaviour of interactive shells_ is to send signals to the whole PG, I really wouldn't worry about process hierarchies and signal order. I have *yet* to find a UNIX command, other than a shell, that actually passes on signals to child processes! Try running `printf '10\n10\n' | xargs sleep` and killing `xargs`, you'll observe that the `sleep` commands are _not killed_, because `xarg` doesn't pass on signals (all [signal handling in `xargs`](https://github.com/aixoss/findutils/blob/33a40be76056c57ad8808c7d007f4b95c727b16b/xargs/xargs.c#L727-L728) is aimed at _listening_ for child process changes). Nor does `find` (only [listens](https://github.com/aixoss/findutils/blob/33a40be76056c57ad8808c7d007f4b95c727b16b/find/pred.c#L1945-L1949) or [`watch`](https://gitlab.com/procps-ng/procps/-/blob/newlib/src/watch.c#L339-350). **Bottom line**: UNIX tools expect job control (through PGs) to handle process tree signalling. Passing on signals is only needed when you explicitly remove control from the interactive shell parent, usually by using separate PGs (the other method is to create a new _session leader_, which really is just a new hierarchy of PGs; this can be done in a script with the [`setsid` command](https://man7.org/linux/man-pages/man1/setsid.1.html)).

mjpieters:
This makes it vastly easier to handle task singnalling; signals can be sent to all
processes in the group, including SIGTERM.

The Apple and Linux variants have largely been combined into a single codebase,
and tests have been enabled for MacOS as a result.

Fixes #372

## Checklist


mjpieters:
I'm planning to look over the Windows-specific code and see what, if anything, can be updated to use command-group as well.

I'll also need to add a changelog entry. :-)

mjpieters:
- The windows build predictably fails because it isn't yet expecting `GroupChild` instead of `Child`
- The iOS build fails because the darwin-procfs doesn't target iOS. I am, again, strongly suspecting that the target should have been `aarch64-apple-darwin` and not `aarch64-apple-ios`, however.

mjpieters:
I went out on a limb and added a commit that replaces iOS with M1/M1Pro builds. That makes all the tests pass.

mjpieters:
Well, except for a flakey test. I guess we need to add in a short sleep somewhere?

mjpieters:
I noticed I created the PR against main, not development. I'll close it and recreate it instead.

mjpieters:
Replaced by #376

PR #376, opened 2022-10-14
Give each task a separate process group

### LINKED ISSUE #372: FR: give child processes a group ID, then use that to kill the process tree

mjpieters:
This makes it vastly easier to handle task singnalling; signals can be sent to all
processes in the group, including SIGTERM.

The Apple and Linux variants have largely been combined into a single codebase,
and tests have been enabled for MacOS as a result.

Fixes #372

## Checklist


mjpieters:
Current status: The client tests fail, because on Mac the tempdir is a (much) longer base path than the path on Linux. This means the status output columns are wider, and the snapshots don't match purely on inter-column whitespace and the border lines being longer. We'll need some method of canonicalising the output.

mjpieters:
Tests are all passing, on Linux and on MacOS. :-)

mjpieters:
I am running out of ideas as to why `daemon::edit::test_edit_flow` flakes out from time to time.

There is nothing about the code involved that's specific to MacOS, nor is the task state incorrect until the very end after the edit flow completed.

I can't even reproduce the issue locally. I'll keep digging whenever I have time, however.

mjpieters:
Status: I can reproduce the flakey test failing with

    ```sh
    while cargo test -q \
        --test daemon_tests daemon::edit::test_edit_flow
    do :; done
    ```

So I can now start tracing the issue.

I've fixed the other flakey tests I found, and using cargo nextest on CI test and coverage runs keeps the builds sane. Adding a GH test reporter action will help surface future flakey tests (nestest can output junit reports).

mjpieters:
Bingo, found the culprit: `tests::helper::wait::wait_for_group_status()` didn't do what it said on the tin, it didn't actually verify if a group had changed to the expected status, so the `test_edit_flow` could end up executing the task edit flow before the groups were paused.

With that, my claim that all tests are passing on both Linux and MacOS is finally actually true for all the tests, all the time! :-D Until the next flakey test found, of course, but a constant loop of tests has not yet shown any more.

mjpieters:
All flaking test related commits have been moved to #378, to keep this PR focused on just process groups and MacOS support.

lib/src/process_helper/apple.rs
```diff
@@ -1,166 +1,19 @@
...
-}
-
-/// This is a helper function to safely kill a child process.
-/// Its purpose is to properly kill all processes and prevent any dangling processes.
-pub fn kill_child(task_id: usize, child: &mut Child, _kill_children: bool) -> bool {
-    match child.kill() {
-        Err(_) => {
-            debug!("Task {task_id} has already finished by itself");
-            false
+pub fn get_process_group_pids(pgrp: libc::pid_t) -> Vec<libc::pid_t> {
```
Nukesor, line 6: From what I can see, this is the only place `libc` is used for.
The `libc::pid_t` type a type alias to `i32` for both linux and darwin systems.
https://docs.rs/libc/latest/libc/type.pid_t.html

Couldn't we just go with `i32` and safe the explicit `libc` dependency?
If it cannot be avoided, this should also be exposed via one of our other dependencies.

I don't feel comfortable to directly work with `libc` when we're using abstraction libraries for the rest of our system interaction. This will only lead to further dependency bloat and incompatibilities in the long run.

lib/Cargo.toml
```diff
@@ -60,10 +61,15 @@ winapi = { version = "0.3", features = [
 ] }
 
 # Unix
-[target.'cfg(not(windows))'.dependencies]
-nix = "0.24"
+[target.'cfg(unix)'.dependencies]
 whoami = "1"
+libc = "0.2.135"
```
Nukesor, line 66: Libc dependency over here.

lib/src/process_helper/linux.rs
```diff
@@ -1,267 +1,11 @@
...
-/// This makes things a bit more complicated, since we have to find out whether a shell is spawned
-/// or not. If a shell is spawned, we do the **Normal case** handling.
-///
-/// If **no** shell is spawned, we have to send the signal to the top-level process only.
-///
-/// If the user also want's to send the signal to all child processes of the task,
-/// we have to get all child-processes of that `$command` process. and send them the signal.
-///
-/// Returns `Ok(true)`, if everything went alright
-/// Returns `Ok(false)`, if the process went away while we tried to send the signal.
```
Nukesor, line 95: You removed all of this logic, but this problem sadly doesn't go away.

For example when executing multiple commands after another `sleep 1 && sleep 1`, a root `sh -c` command will be spawned, which the `sleep 1` processes attach to.

For single stand-alone processes however, `sh -c` directly replaces itself with the target command.
Shells don't forward signals by default, which means that signals won't be propagated as expected.

For the purpose of convenience, I did all of this logic to spare the user from having to add a `-c` flag depending on whether they run a something that will result in a `sh -c` process or not.

I'm very much in favor of re-adding this behavior.

Nukesor (changes requested):
Thanks for this MR!

It's pretty massive and it's awesome that you managed to implement the apple process handling!

Still, I have a few concerns with the approach on process signal handling.
Generally I think that ProcessGroups are a great idea, especially for `SigKill` signals, since processes won't get a chance to propagate the signal in the first place.

I'm going to give you a short rundown on the reasoning behind they way I built this whole stuff:

## Unix signals

In general, Unix signals are supposed to be handled in a hierarchic way.
Parent processes get the signal and they then decide what to do with their child processes.

Every program that spawns child processes should have a proper signal handler in place and handle their child processes according to their signals.
For instance, pueue does this.
Once it receives a Sigterm/Sigint, it enters a graceful shutdown mode, kills all child processes, cleans them up and then continues to shut down.

This mode of operation is important, as quite a few programs depend on this.
From personal experience, this is especially the case in scientific environments, where multi-process programs are still quite common.

## `sh -c` detection

On Linux, I decided to use the oldschool `posix` shell, as this is the only shell we can be sure about.
The posix shell, as well as other common shells such as zsh and bash, replace the actual shell process if there's no shell specific logic.

I.e. A simple `sh -c 'sleep 10` will become a stand-alone `sleep 10` process.

A more complex `sh -c `sleep 10 && sleep 10` however, will look like this:

```
|
 - sh -c 'sleep 10 && sleep 10'
 |
  - sleep 10
```

The problem is, that most shells don't propagate Unix signals by default. This must be explicitly done in a shell script.

For this reason, my old logic tried to detect whether there was a `sh -c` root process and send the signal to the `sleep 10` process instead.
The `sh -c` process is a mere implementation detail and there's no reason to expose this logic to the user.
They should only worry about the process structure of the inner command.

## Bash script convenience

The main reason for adding the `-c` flag, was to give the user a simple escape hatch out of their shell scripts.
Users were writing shell scripts, but they didn't care about Unix signal propagation.

For this reason, I added the `-c` flag, so they don't have to care about Unix signal propagation, but could rather let Pueue handle it.

## Concerns

This is a super tricky topic.
Most people don't care about Unix signals and they don't want to implement handling of such signals.

However, some programs very much depend on this signal hierarchy and sending signals to nested child processes in the wrong order could lead to a broken state, bugs and collateral damage (corrupted files, broken I/O pipelines and such).

As I said at the beginning, I see how this is definitely the right choice for actually killing (SigKill) process groups.
It's probably even the desired approach for other Signals in 99% of all use cases, but for the remaining 1% this could get quite ugly.

I'm not sure if I'm overly cautious about this, but it feels like this breaks the idea behind hierarchic Unix signal handling.

mjpieters:
I hear your concerns! I should probably have updated the PR description when my own understanding of Process Groups (PGs) matured, but PGs are how the [_shell_ handles job control](https://en.wikipedia.org/wiki/Job_control_(Unix)#Implementation), when running interactively. The core reason for this PR is that _in normal shell operations_, it is the use of PGs that make it possible to exit a hierarchy of processes. I'll dive a little into how this works below. To keep terminology simple, lets call that the 'interactive shell', and any non-interactive shell process a 'shell script'.

So, in an interactive shell, when you start a command, be that a shell script or something else, the interactive shell creates a new PG for that command. From there on out, it controls that PG using signals, e.g. when you put the command in the background (with `Ctrl-Z` or by starting it in the background with `&`), killing that command is done by sending `SIGKILL` to the whole PG. If the interactive shell exits (say, because something send it a SIGKILL signal), it'll signal each PG it started with `SIGHUP`, wait for those PGs to exit, before exiting itself.

It is only when you _don't_ run a shell interactively that this doesn't happen. That's because that would interfere with the normal operation of a parent (interactive) shell, you don't want the process tree started by a shell script to start a new PG as that would remove control from the interactive shell.

Note: It is actually _surprisingly hard_ to handle signals in a bash script, because _unless you put commands it starts in the background_, a shell script [**won't handle signal hooks set with `trap` until any active child process exits**](https://tldp.org/LDP/Bash-Beginners-Guide/html/sect_12_02.html#sect_12_02_02). You _can_ enable 'monitor mode' in a shell script with `set -m`, in which mode _background_ processes started by the shell script will be given a new PG, but this doesn't happen for foreground processes. This is why you can't just kill the parent shell script and hope that the child processes come along. In fact, monitoring mode (`set -m`) is mostly used to [prevent the interactive shell from killing a child background process when the interactive shell exits](https://unix.stackexchange.com/questions/196603/can-someone-explain-in-detail-what-set-m-does).

My own understanding of PG use in interactive shells has evolved over the course of creating this PR, and in retrospective it may have been a mistake to retain the `-c / --children` behaviour. I think it should be the _default_, in that all signals sent by pueue to a child process should, by default, go to the whole PG, just like [standard interactive shell job control](https://unix.stackexchange.com/questions/594141/what-determines-whether-a-scripts-background-processes-get-a-terminals-sigint), where `kill %jobid` will [send the signal to the PG](https://github.com/bminor/bash/blob/f3a35a2d601a55f337f8ca02a541f8c033682247/builtins/kill.def#L232-L248).

On process hierarchies: Given that _the default behaviour of interactive shells_ is to send signals to the whole PG, I really wouldn't worry about process hierarchies and signal order. I have *yet* to find a UNIX command, other than a shell, that actually passes on signals to child processes! Try running `printf '10\n10\n' | xargs sleep` and killing `xargs`, you'll observe that the `sleep` commands are _not killed_, because `xarg` doesn't pass on signals (all [signal handling in `xargs`](https://github.com/aixoss/findutils/blob/33a40be76056c57ad8808c7d007f4b95c727b16b/xargs/xargs.c#L727-L728) is aimed at _listening_ for child process changes). Nor does `find` (only [listens](https://github.com/aixoss/findutils/blob/33a40be76056c57ad8808c7d007f4b95c727b16b/find/pred.c#L1945-L1949) or [`watch`](https://gitlab.com/procps-ng/procps/-/blob/newlib/src/watch.c#L339-350). Passing on signals is only needed when you explicitly remove control from the interactive shell parent, usually by using separate PGs (the other method is to create a new _session leader_, which really is just a new hierarchy of PGs; this can be done in a script with the [`setsid` command](https://man7.org/linux/man-pages/man1/setsid.1.html)).

The more I reread what I wrote here, the more I am convinced that I should make `--children` the default (deprecating the switch) and instead add an option to limit the signal to the initial process, only. I'm not sure that I personally would ever have a use-case for the latter switch, though.

At any rate, PGs make the need to detect whether or not `sh -c` started new processes or just replaced itself entirely moot. Kill the PG, you kill whatever processes you started.

daemon/task_handler/callback.rs
```diff
@@ -25,7 +25,7 @@ impl TaskHandler {
         let mut command = compile_shell_command(&callback_command);
 
         // Spawn the callback subprocess and log if it fails.
-        let spawn_result = command.spawn();
+        let spawn_result = command.group_spawn();
```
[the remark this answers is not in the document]
mjpieters: Yeah, I was being _too_ consistent here.

tests/helper/daemon.rs
```diff
@@ -70,7 +69,7 @@ pub async fn wait_for_shutdown(pid: i32) -> Result<()> {
 
     while current_try < tries {
         // Process is still alive, wait a little longer
-        if process.is_alive() {
```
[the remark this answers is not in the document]
mjpieters: Darn, I didn't fully grasp that `is_alive()` does more than just check that the PID exists.

We can use [`try_wait()`](https://doc.rust-lang.org/std/process/struct.Child.html#method.try_wait) here then, as that'll reap the status of any zombie process process. (There is no external 'dead' state for processes, that's an internal Linux process state used while the kernel is cleaning up the process entries).

tests/helper/client.rs
```diff
@@ -178,3 +171,105 @@ pub fn assert_stdout_matches(
...
+                        *table_started = true;
+                        // Once we are inside a table, trim the path column by cutting out the characters
+                        // at [from, until) (taking into account how much the line has already shrunk due
+                        // to replacements)
+                        let until = until - trimmed;
+                        let chars = line.chars();
+                        Some(chars.clone().take(from).chain(chars.skip(until)).collect())
+                    }
+                }
+            },
```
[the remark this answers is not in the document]
mjpieters: I agree with on not liking this, but I must note that you'd have run into the same problem elsewhere too as even on Linux there are no guarantees about the temporary directory length. The tests (rightly) use `tempfile::TempDir` which in turn bases the path on the `TMPDIR` environment variable, and that's not guaranteed to be `/tmp` in every environment.

Using separate templates for different OSes would also create a huge maintenance burden, more likely to lead to test suites for platforms you don't have direct access to to go stale rather than receive test suite updates and / or put up road-blocks for outside contributors. If ever you'd expand the test suite coverage to include Windows you'd *again* run into this exact same problem as there the temporary directory is located in `%userprofile%\AppData\Local\Temp`, and so the path length there depends on the length of the username. This then can lead to tests failing for some users because their username happens to be shorter or longer than on the GitHub Windows runners. So in my view, *given the nature of the client table generator and the variability of path name lengths*, dynamically adjusting the column width is a necessity. The alternative would be to parse tables in client output into some kind of data structure and then either re-render to match a template, or template the expected table contents in a similar data structure.

Note that what you are using is very much **not** snapshot testing. Real snapshots are managed entirely automatically; you either have a snapshot and the test output must match entirely, or you explicitly capture a new snapshot to compare in future tests. That only works if you can remove all the variation between tests, e.g. by using mocks for paths, system time, etc. What you have instead is a templating system to produce expected outputs. Switching to real snapshots is also an option, but would be a lot of work to retool the test suite to use mocks.

lib/src/process_helper/mod.rs
```diff
@@ -21,6 +32,15 @@ pub use self::apple::*;
...
 pub use self::windows::*;
+#[cfg(target_os = "windows")]
+/// Shim signal enum to simplify cross-platform handling
+pub enum Signal {
+    SIGINT,
+    SIGKILL,
+    SIGTERM,
+    SIGCONT,
+    SIGSTOP,
+}
```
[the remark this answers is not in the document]
mjpieters: Sure, I'll move it.

lib/src/process_helper/apple.rs
mjpieters: You are quite right, it's i32 on either Linux or Darwin. Only the `darwin_libproc` crate actually uses `libc::pid_t` here whereas the `procfs`  just uses `i32`, and so I should have looked at what `libc::pid_t` is defined as on darwin.

mjpieters:
For what it's worth: I've been running my pipelines with this version for over a week now. I updated the version Cargo.toml to 2.2.0-alpha.0, and `lib/Cargo.toml` to `0.22.0-alpha.0`, ran `cargo update -w`, and built a release version with `cargo build --release --lock`. It works so fantastically well, that I _really don't want to go back now_.

I have a lot of jobs with non-trivial process trees. They are usually shell scripts that do non-trivial things, all the way up to GNU parallel jobs which farm out to other servers. Before I prefixed my commands with `echo "PGID=$(ps --format pgid= --pid $$ | tr -d " ")"`, then wrapped the whole thing in `setsid --wait bash -c`, just so I could kill or otherwise signal the process group created with that. This did mean that `pueue pause` **won't work anymore** as that would only signal the `setsid` session leader process, and if the pueue _daemon_ died all my tasks would be left to run as children of the init process. Mostly, that was already the case for any task using more than 2 levels of processes deep, which is about all of them. But at least I could grep for `PGID` and use `pkill -g` on that to kill or signal the task processes properly.

Now, I just use `pueue kill`, **and it just works**. So do pueue pause / pueue start. Shutting down the daemon takes all the task processes with it. It's wonderful!

Is there any way we can get this out to more people to test, as a pre-release or alpha test group? Do you have any examples of workloads where the process tree hierarchy is important, where you _have_ to kill just a top-most process and let it handle signals from there?

Nukesor:
First up, thanks for the detailed explanation and the linked resources! This made researching this topic a lot more enjoyable.

I agree that using process groups by default seems like a reasonable default. To be honest, I didn't expect shells to already use process groups, which is why I never checked them.

I'm still a bit confused about the default behavior though.
I tried to reproduce the behavior explained in the [Signal inteception? propagation? No](https://unix.stackexchange.com/a/594236) SO answer, but I'm getting different results.

If one run a shell script `test.sh` with this content:
```sh
#!/bin/bash
set -m
sleep 60 & sleep 60
```

```sh
./test.sh
# Run this in a separate shell:
# ps xao pid,ppid,pgid,sid,comm | rg sleep
#  14594   14593   14594   12462 sleep
#  14595   14593   14595   12462 sleep
# Press CTRL+C
ps xao pid,ppid,pgid,sid,comm | rg sleep
14594       1   14594   12462 sleep
```

This shows, that a CTRL+C on a shell script doesn't send the signal to all background tasks.
A single sleep (the one in the background) doesn't get the signal and is attached to the root ppid `1`.

The same happens with the `set +m` option.


**Anyhow**
I agree that it's the most convenient thing to just send the signal to the whole process group! It covers the 90+% usecase and those should always be the default.
We could leave the option to send the signal to only the parent out for now. This should be easy to re-add once people come up and we'll see some proper use-cases in the process.

Still, this is somewhat of a breaking change. I'm currently thinking about how to properly transition from the current behavior.

A new major version is definitely in order, we should keep the `-c` flag and add a warning that this is deprecated and by default the case via process groups.

tests/helper/client.rs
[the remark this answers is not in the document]
Nukesor: I agree that the current snapshot testing only works because all templated outputs have the exact same format.
I.e. all paths are located in `/tmp` with a 8 char prefix. Or system time, that's formatted in a very specific way that always has the same length.

Though I got an idea on how we can do this properly!
The new `status` query syntax I've recently added would allow us to simply hide all columns that aren't of any interest to us.

This could completely remove the need for templating.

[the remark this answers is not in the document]
Nukesor: Let's stick with the current approach for now and investigate a proper solution to this in a follow-up MR!

lib/Cargo.toml
```diff
@@ -60,10 +61,14 @@ winapi = { version = "0.3", features = [
...
-[target.'cfg(any(target_os = "linux", target_os = "freebsd"))'.dependencies]
-procfs = { version = "0.14", default-features = false }
+# Linux only
+[target.'cfg(target_os = "linux")'.dependencies]
+procfs = "0.14.1"
+
+# Apple only
+[target.'cfg(target_vendor = "apple")'.dependencies]
+# fork that fixes a version conflict, see heim-rs/darwin-libproc#3
+darwin-libproc = { version="0.2.0", git = "https://github.com/jamwaffles/darwin-libproc.git", rev = "cbd571e4be87a6603de328d1f1f0ad54d7bdfedf" }
```
Nukesor, line 74: This might actually be a blocker for publishing a new release.
One isn't allowed to publish packages to crates.io that depend on git repositories.

As far as I know, the version `0.2.0` will then be used as a fallback, effectlively breaking this release for all apple platforms.

Nukesor: The darwin-libproc library also seems quite abandoned.

Do you think there's any chance on this getting fixed in time?

tests/helper/client.rs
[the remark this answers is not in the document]
Nukesor: Just did a little POC for this approach:
https://github.com/Nukesor/pueue/commits/simplified_snapshot_testing_poc

We could take this approach and even keep everything except the path, which is currently the only thing that's platform specific. The rest of the formatting logic is controlled by Pueue and thereby deterministic.

lib/Cargo.toml
mjpieters: 🤦 I'll see if https://github.com/andrewdavidmackenzie/libproc-rs can replace it. I was under the impression that heim was an active project, and darwin-libproc is part of heim. Turns out I was wrong about that :-(

tests/helper/client.rs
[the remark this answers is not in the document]
mjpieters: Interesting approach! Do you feel that that meets your testing goals, what might be missed if the paths are not part of these tests?

Some further thoughts on snapshotting: Snapshotting is a way to test a UI. In the case of the pueue command line, that's the stdout or stderr output that the command produces. You usually **separate UI testing from behaviour testing**, in that you finely control the variables that affect the UI layer and so can control the output. "If color is enabled, given a task table in this state, the output should match this snapshot", and "given a task that is still running, verify the output of a table with the end column enabled against a snapshot", etc.

Your tests don't isolate the UI from how the input is produced, they are actually integration tests, as networking library, the daemon, and the OS running task processes are all involved. I'd actually use the JSON output option to verify behaviour in such an integration test, at which point you'd not have to worry about column widths being variable. And then test `client::display`, in isolation, with snapshots. For the integration tests, the goal is to test how the system as a whole operates, and the exact formatting of the output matters far less there, and you can cover all bases for the table-formatted output much, much more easily when driving the client::display module directly.

lib/Cargo.toml
mjpieters: libproc-rc is missing support for enumerating processes in a process group, see andrewdavidmackenzie/libproc-rs/issues/85 for my request that that be added.

mjpieters: Actually, this is a **test** dependency only now. I'll rework it as such.

>  One isn't allowed to publish packages to crates.io that depend on git repositories.

Can you publish a package where only the test dependency is a git repo?

mjpieters: I found confirmation that git dev-dependencies are ignored, so I can make this work. Force push incoming.

Nukesor:
Thanks again for the thourough explanation! I'll find a bit more time to do some in-depth research on this in the next few weeks :)

An explicit warning would be awesome! This should probably be a normal `println!()`, as `warn!` level messages won't be displayed by default, as the default log level is `ERROR`.

mjpieters:
> An explicit warning would be awesome! This should probably be a normal `println!()`, as `warn!` level messages won't be displayed by default, as the default log level is `ERROR`.

You ask, you receive! See [these lines in `main.rs`](https://github.com/Nukesor/pueue/blob/7fc98c784fbcad7e2597a3fc28eddfd4c83a2ca0/client/main.rs#L87-L101).

Nukesor:
Ok.

I was confused by the statement that all processes in the same process group of the terminal receive CTRL+C signals. However, this doesn't seem to be the case for background tasks, even if they are in the same process group.

It's even more confusing that the job control bash docs have this statement:
`Background processes are those whose process group ID differs from the terminal’s; such processes are immune to keyboard-generated signals.`
But it's possible to have background tasks that're in the same process group as the terminal.

I'll give this a final review round, merge it and prepare the v3.0.0 release candidate :).

tests/helper/client.rs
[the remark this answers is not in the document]
Nukesor: I agree that these are mostly integration tests or even 'end-to-end' tests.

I'm usually a huge fan of having mostly integration tests, as they are the closest thing to real world scenarios and thereby tend to catch some scenarios one wouldn't normally think about when testing individual pieces of logic.
In the case of most client tests, my goal was to get a good coverage of the commandline usage and it's output.

I won't try to get a 100% coverage on the output formatting code, but rather a "good enough" coverage.
The output comparison is mostly there to check whether the client worked as expected.

However you're completely right and there are a few tests, such as the color test and the query tests where we actually test the client formatting/display code. The quality of those tests will definitely be degraded by the removal of the `path` from the status table, but it feels like this a okayish compromise for now.
Anyhow, we could definitely think about converting those specific tests to something that's more of a unit test.

Until then, we can still verify any path related logic via the `--json` output, as we're doing in the client edit and restart tests right now.

I created a ticket for this refactoring :)
https://github.com/Nukesor/pueue/issues/390

Nukesor (approved):
Awesome :)

This looks good to me! Thanks again for your work, research and patience when working on this huge change!

PR #378, opened 2022-10-17
Address flaky tests

### LINKED ISSUE #379: [BUG] message handler doesn't verify if task handler has processed message
#### Describe the bug

When the daemon message handler receives a message that requires the task handler to do something (like start a process), the message is forwarded via a channel, as the task handler runs in a separate thread.

At no point is there any verification that the message has been handled by the task handler, before the message handler decides what response to give to the handled message.

This can lead to weird outcomes, where the message handler states that starting a task failed, but the task handler _later on_ starts the task anyway.

#### Steps to reproduce the bug

Check out my Flaky Test PR, #378, or apply the changes from the *Dev support: add test-log to help debug tests* commit in that PR to add `test-log` to the project + the `test(...)` attribute update for the `daemon::edit::test_edit_flow` test.

Run the test in a tight loop:

```sh
while RUST_LOG=debug cargo nextest run --workspace --hide-progress-bar --status-level=fail \
    daemon::edit::test_edit_flow
do :; done
```

At some point the test fails, but logs will show that the *Start* message failed because the task is locked for editing:

```
[2022-10-17T16:49:20Z DEBUG pueue_lib::network::protocol] Received message: Start(
        StartMessage {
            tasks: TaskIds(
                [
                    0,
                ],
            ),
            children: false,
        },
    )
[2022-10-17T16:49:20Z DEBUG pueue_lib::network::protocol] Sending message: Failure(
        "The command failed for tasks: 0",
    )
```

Between these to log entries, the message handler has sent the start message on to the task handler, and *immediately* produces a response from the current state.

In the case of a test failure, an `Edit` message is received by the daemon and also handled, putting the task into an unlocked, queued state.

However, the task handler has not yet obtained the mutex lock during all of this time. Even though the edit flow has been completed, the task handler finally can lock the state, finds the task in a queued state and so starts the task:

```
[2022-10-17T16:49:20Z INFO  pueue_daemon_lib::task_handler::spawn_task] Started task: ls -ahl
```

#### Expected behavior

The message handler should only produce a response once the task handler has processed the message too, otherwise it can produce responses that are not related to the actual state of the task handler.

#### Additional context

I can reproduce this issue on my Macbook Pro with M1 Pro CPU, I've not yet seen it when I run the test in a loop in a docker container running Linux. There is nothing in the implementation from preventing this bug from cropping up in Linux, however.

mjpieters:
Handle flaking tests better at the CI level, make it easier to find flaking tests and address some of those issues.

This PR extracts work done to mitigate flaky tests from #376. I'll rebase that PR on top of this one (so include the commits here into that PR). If this PR is merged before the other, the other can be rebased again trivially.

Fixes #379.

## Background

Some tests in the test suite flake, fail based on exact timings or because of issues with test helpers. These failures are not signs that the project code is broken and so should not block accepting new work.

This PR aims to update how tests are run on GitHub Actions so flaky tests can be detected and be re-run to verify that the code-under-test is not the source of the test failure.

## Changes included in the PR

1. Switch to [cargo nextest](https://nexte.st/) to run the test suite. Not only does nextest run the test suite faster, it also can re-run tests that failed to verify that the failure wasn't a flake.

2. Adds a [test reporter action](https://github.com/marketplace/actions/publish-test-results), which surfaces test results in PRs and GitHub commit contexts. This makes it possible to see that a test flaked, as flaky tests, ideally, are fixed so they can't flake.

  **Note**: This part requires an additional GitHub workflow to actually publish the test report. This workflow will *not run* while still confined in this PR, because GitHub will not yet be able to find the workflow in the default
  (main) branch until it is merged into *that branch*. I've created a [demonstration PR](https://github.com/mjpieters/pueue/pull/2) in my own fork that shows what these reports look like.

3. Update dev-dependencies to make it easier to debug flaking tests, by making
   it trivially easy to enable logging for specific tests. I have found the
   existing logging already present in the pueue codebase to be _excellent_,
   and now only a `RUST_LOG=debug` environment variable away from becoming
   visible when a test fails.

4. And finally, individual commits to fix specific flaking tests I found I was
   able to fix.

## How to find and debug flaking tests

Run the test suite in a continuous loop; preferably with cargo nextest as that's faster; I've tuned the command-line switches to minimise output for runs without test failures:

```sh
while cargo nextest run --workspace --hide-progress-bar --status-level=fail
do :; done
```

You can do the same with the default `cargo test` command:

```sh
while cargo test --quiet
do :; done
```

Once a flaking test has been identified, add a test filter to focus on that specific test, add `RUST_LOG=debug` to the command-line to ensure you get plenty of log output when a failure happens, and if the test in question hasn't already been updated to [use the test-log macro](https://crates.io/crates/test-log), do so before trying to reproduce the issue:

```sh
while RUST_LOG=debug cargo nextest run --workspace --hide-progress-bar --status-level=fail \
    daemon::edit::test_edit_flow
do :; done
```

The cargo test equivalent is:

```sh
while RUST_LOG=debug cargo test --workspace --quiet \
    --test daemon_tests daemon::edit::test_edit_flow
do :; done
```

When you next hit a failure, you should have a plethora of information as to what pueue was doing up to that point.

## Checklist

### DIFF
--- a/.github/workflows/test-report.yml
+++ b/.github/workflows/test-report.yml
@@ -1,0 +1,71 @@
+# This workflow makes it possible to publish test reports without running into
+# permission issues when the test workflow was run from a fork or by Dependabot.
+#
+# The test workflow uploads a junit file per matrix target as an artifact, plus
+# the worflow events file, both of which this worfklow buids upon. Note that
+# the events file artifact, specifically, is expected to be named 'Event File'.
+#
+# See the [Publish Test Results action documentation][ptr] for more information.
+#
+# [ptr]: https://github.com/marketplace/actions/publish-test-results#support-fork-repositories-and-dependabot-branches
+
+name: 'Test Report'
+on:
+  workflow_run:
+    workflows: ['Test Build']
+    types:
+      - completed
+
+permissions: {}
+
+jobs:
+  test-results:
+    name: Test Results
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.conclusion != 'skipped'
+
+    permissions:
+      checks: write
+
+      # permission to comment on PRs
+      pull-requests: write
+
+      # permission to download artifacts
+      actions: read
+
+    steps:
+      - name: Download and extract artifacts
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+        run: |
+            # Unzip all artifacts created by the triggering workflow into
+            # directories under an `artifacts/` directory.
+            # 
+            # This uses `gh api` to output the name and URL for each artifact as
+            # tab-separated lines, then uses `read` to take each name and URL
+            # and download those to named zip files, and finally extracting those
+            # zip files into directories with matching names.
+
+            mkdir -p artifacts && cd artifacts
+
+            # The artifacts URL from the *triggering* test workflow, not *this*
+            # workflow.
+            artifacts_url=${{ github.event.workflow_run.artifacts_url }}
+
+            gh api "$artifacts_url" -q '.artifacts[] | [.name, .archive_download_url] | @tsv' | while read artifact
+            do
+                IFS=$'\t' read name url <<< "$artifact"
+                gh api $url > "$name.zip"
+                unzip -d "$name" "$name.zip"
+            done
+
+      # Run the publisher. Note that it is given the 'Event File' artifact
+      # created by the test workflow so it has the *original* webhook payload
+      # to base its context on.
+      - name: Publish Test Results
+        uses: EnricoMi/publish-unit-test-result-action@v2
+        with:
+          commit: ${{ github.event.workflow_run.head_sha }}
+          event_file: artifacts/Event File/event.json
+          event_name: ${{ github.event.workflow_run.event }}
+          junit_files: "artifacts/**/*.xml"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -75,6 +75,9 @@ jobs:
           components: llvm-tools-preview
           target: ${{ matrix.target }}
 
+      - name: Install cargo-nextest
+        uses: taiki-e/install-action@nextest
+
       - uses: actions/cache@v3
         with:
           path: |
@@ -97,7 +100,9 @@ jobs:
       - name: cargo test
         uses: actions-rs/cargo@v1
         if: ${{ !matrix.cross }}
+        env:
+          NEXTEST_PROFILE: ci   # defined in .config/nextest.toml
         with:
-          command: test
-          args:  --workspace --target=${{ matrix.target }}
+          command: nextest
+          args:  run --workspace --target=${{ matrix.target }}
           use-cross: ${{ matrix.cross }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: Test build
+name: 'Test Build'
 
 on:
   push:
@@ -106,3 +106,24 @@ jobs:
           command: nextest
           args:  run --workspace --target=${{ matrix.target }}
           use-cross: ${{ matrix.cross }}
+
+      # The test result artifacts are used by the test-report.yaml workflow.
+      - name: upload test results
+        uses: actions/upload-artifact@v3
+        if: ${{ !matrix.cross }}
+        with:
+          name: Test results (${{ matrix.target }})
+          path: target/nextest/ci/junit.xml
+
+  # the event file (containing the JSON payload for the webhook triggering this
+  # workflow) is needed to generate test result reports with the correct
+  # context. See the test-report.yaml workflow for details.
+  event_file:
+    name: "Event File"
+    runs-on: ubuntu-latest
+    steps:
+    - name: Upload
+      uses: actions/upload-artifact@v3
+      with:
+        name: Event File
+        path: ${{ github.event_path }}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,6 +89,13 @@ rstest = "0.15"
 serde_yaml = "0.9"
 similar-asserts = "1"
 
+# Make it easy to view log output for select tests.
+# Set log level for tests with RUST_LOG=<level>, use with failed tests or
+# disable test stdout/stderr capture (`cargo test -- --nocapture` / `cargo
+# nextest run --no-capture`)
+env_logger = "0.9.1"
+test-log = "0.2.11"
+
 # Test specific dev-dependencies
 [target.'cfg(any(target_os = "linux", target_os = "freebsd"))'.dependencies]
 nix = "0.24"
--- a/tests/daemon/edit.rs
+++ b/tests/daemon/edit.rs
@@ -1,6 +1,7 @@
 use std::path::PathBuf;
 
 use anyhow::{bail, Result};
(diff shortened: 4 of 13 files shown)


mjpieters:
The only check failure here, raised by codecov, will most likely go away if #382 were to be merged (and I rebase this PR).

.github/workflows/test.yml
```diff
@@ -97,7 +100,26 @@ jobs:
...
+
+  event_file:
+    name: "Event File"
+    runs-on: ubuntu-latest
+    steps:
+    - name: Upload
+      uses: actions/upload-artifact@v2
+      with:
+        name: Event File
+        path: ${{ github.event_path }}
```
Nukesor, line 125: If i understand this correctly, this uploads the full webhook event, that would be send if there were any webhooks specified?
This is then unpacked in the test-report action and used in the publish-unit-test-result-action call.

It would be great to have a few lines of docs on why this is desirable.

tests/daemon/edit.rs
```diff
@@ -26,7 +27,7 @@ async fn create_edited_task(shared: &Shared) -> Result<EditResponseMessage> {
     }
 }
 
-#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+#[test(tokio::test(flavor = "multi_thread", worker_threads = 2))]
```
Nukesor, line 30: Would you recommend to leave this in here?
If so, should we consider adding this to all tests?

Cargo.toml
```diff
@@ -89,6 +89,13 @@ rstest = "0.15"
 serde_yaml = "0.9"
 similar-asserts = "1"
 
+# Make it easy to view log output for select tests.
+# Set log level for tests with RUST_LOG=<level>, use with failed tests or
+# disable test stdout/stderr capture (`cargo test -- --nocapture` / `cargo
+# nextest run --no-capture`)
```
Nukesor, line 95: Maybe we should start a new document or a wiki page for development practices.
Your tutorial on how to catch and debug flaky tests would also be good to have in there :)

Nukesor (changes requested):
Nice.

Overall this looks good to me :) Thanks for tackling those flaky tests.
I didn't expect to have that many flaky tests left in this codebase :sweat_smile: 

Just gave nextest a spin and it looks great! Especially the flakiness detection is a godsend. I'll probably introduduce this i other projects as well!

.github/workflows/test-report.yml
```diff
@@ -0,0 +1,47 @@
+name: 'Test Report'
+on:
+  workflow_run:
+    workflows: ['Test Build']
+    types:
+      - completed
```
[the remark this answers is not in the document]
mjpieters: This allows the report runner to create a report even if the tests were run against a forked repo (e.g. any PR from someone other than you).

See the [documentation for the report action](https://github.com/marketplace/actions/publish-test-results#support-fork-repositories-and-dependabot-branches); the separate workflow is run whenever the test workflow completes, but it is run within the pueue repo, always. The downside is that it must be present in the default branch for GitHub to find it.

.github/workflows/test-report.yml
```diff
@@ -0,0 +1,47 @@
...
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+        run: |
+           mkdir -p artifacts && cd artifacts
+
+           artifacts_url=${{ github.event.workflow_run.artifacts_url }}
+
+           gh api "$artifacts_url" -q '.artifacts[] | [.name, .archive_download_url] | @tsv' | while read artifact
+           do
+             IFS=$'\t' read name url <<< "$artifact"
```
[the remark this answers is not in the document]
mjpieters: Sorry, it's a straight copy-pasta from the [Publish Test Results documentation](https://github.com/marketplace/actions/publish-test-results#support-fork-repositories-and-dependabot-branches).

Happy to add a comment or two anyway :-)

.github/workflows/test.yml
```diff
@@ -97,7 +100,26 @@ jobs:
...
+        with:
+          name: Test results (${{ matrix.target }})
+          path: target/nextest/ci/junit.xml
+
+  event_file:
+    name: "Event File"
+    runs-on: ubuntu-latest
+    steps:
+    - name: Upload
+      uses: actions/upload-artifact@v2
```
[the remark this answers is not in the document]
mjpieters: Missed this one, I spotted the other when I copied the sample workflow setup 🤦

tests/daemon/edit.rs
mjpieters: It shouldn't make much difference if left in place. It'll be helpful for future debugging issues.

I recommend adding it to tests where you feel the need to see log output, on a case by case basis. There is little point in adding it to all the tests just for the sake of it.

Cargo.toml
mjpieters: Sounds like a plan! Perhaps add a DEVELOPMENT.md pointing to such a page?

.github/workflows/test.yml
mjpieters: Again, it's a requirement set by the publishing action, as it uses the event file as a basis. The separate workflow handling the reports means it would otherwise have the wrong context.

.github/workflows/test-report.yml
```diff
@@ -0,0 +1,47 @@
...
+
+      # permission to download artifacts
+      actions: read
+
+    steps:
+      - name: Download and extract artifacts
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+        run: |
+           mkdir -p artifacts && cd artifacts
```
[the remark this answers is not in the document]
mjpieters: I suspect that the downloaded would have the wrong context, as the artefacts were uploaded by a different workflow.

The bash script takes the events file info instead. 

This is prescribed by the report action documentation.

Cargo.toml
Nukesor: Sounds good to me!

.github/workflows/test.yml
```diff
@@ -97,7 +100,26 @@ jobs:
...
         with:
-          command: test
-          args:  --workspace --target=${{ matrix.target }}
+          command: nextest
+          args:  run --workspace --target=${{ matrix.target }}
           use-cross: ${{ matrix.cross }}
+
+      - name: upload test results
+        uses: actions/upload-artifact@v3
+        if: ${{ !matrix.cross && ( success() || failure() ) }}
```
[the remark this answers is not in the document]
mjpieters: Looks like I misread when `success()` is `true`, I thought it would be false when the workflow was cancelled. Re-reading the docs it is entirely redundant here.

Cargo.toml
mjpieters: I've made a start here: https://github.com/Nukesor/pueue/wiki/Development

Shall I leave it up to you to fill in a DEVELOPMENT.md file?

Nukesor: I'll extend the wiki entry and transfer it to the DEVELOPMENT.md lateron :)

Nukesor (approved):
Nice. I thinks we are good to merge this :+1: 

There seem to be a few merge conflicts with your other branch. I'll merge once this is done :)

mjpieters:
> There seem to have a few merge conflicts with your other branch though. I'll merge once this is done :)

Rebase done!

PR #380, opened 2022-10-18
Use async sleep

mjpieters:
Use tokio::time::sleep rather than std::thread::sleep; the async version cooperates (yields to other tasks), so _just the task_ is blocked, not all tasks being run in the same thread.

It's a minor efficiency improvement, so I didn't include a changelog entry. LMK if this warrants one anyway.

## Checklist

### DIFF
--- a/client/commands/wait.rs
+++ b/client/commands/wait.rs
@@ -1,10 +1,10 @@
 use std::collections::HashMap;
-use std::thread::sleep;
 use std::time::Duration;
 
 use anyhow::Result;
 use chrono::Local;
 use crossterm::style::{Attribute, Color};
+use tokio::time::sleep;
 
 use pueue_lib::network::protocol::GenericStream;
 use pueue_lib::task::{Task, TaskResult, TaskStatus};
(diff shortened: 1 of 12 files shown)


Nukesor:
The async wait is probably nice for the tests, even though this'll barely noticable.

In the case of the client, it won't make any difference, as there's only a single task on a single thread at all times anyway.
It might even make the client a bit slower, as the executor has to actively continue the yielded task.
I don't see anything against using the async wait though. It's probably nice to use it everywhere for consistency purposes.

Nukesor (approved):

mjpieters:
> The async wait is probably nice for the tests. Though I'm not sure if it'll be noticable :D.
> 
> In the case of the client, it won't make any difference, as there's only a single task on a single thread at all times anyway. It might even make the client a bit slower, as the executor has to actively continue the yielded task. I don't see anything against using the async wait though. It's probably nice to use it everywhere for consistency purposes.

There is little change in the timings in the test suite, the sleeps don't take up that much time compared with everything else.

I may however, address some blocking code in the daemon in future. I think there's some I/O (outside of the task handler) that should really be be handled async.

PR #418, opened 2023-02-18
Add encoding detection and conversion

### LINKED ISSUE #417: [BUG] errors with non-UTF-8 task logs
#### Describe the bug

In a task that outputs multibyte characters on Windows, `pueue log` command throws an error.

#### Steps to reproduce the bug

1. Create a Python script `a.py` with the contents `print('abcあdef')`.
2. Add a task with `pueue add python a.py`. The task will complete successfully.
3. Run `pueue log 0` to view the log, and it shows only `abc` not `abcあdef`.

#### Expected behavior

Complete logs are displayed without errors.

#### Logs/Output

`pueue log` outputs the following:

```
┌───────────────────────────────────┐
│ Task 0:    completed successfully │
└───────────────────────────────────┘
Command: python a.py
   Path: C:\***
  Start: Fri, 17 Feb 2023 22:30:48 +0900
    End: Fri, 17 Feb 2023 22:30:49 +0900

output:
abcFailed reading local log file: Windows stdio in console mode does not support writing non-UTF-8 byte sequences
```

I am using python 3.9.13 but I think it is not a python-related problem.

#### Additional context

- Operating System: Windows 11 Home 64bit
- Pueue: 3.1.1

I think the problem is [here](https://github.com/Nukesor/pueue/blob/main/pueue/src/client/display/log/local.rs#L28). 
As stated in this doc https://doc.rust-lang.org/std/io/fn.stdout.html#note-windows-portability-considerations, `std::io::stdout` does not support non-UTF-8 byte sequences, so it cause errors with outputs encoded in a character code other than UTF-8.

yuma140902:
This PR adds the ability to detect character encoding of task logs and convert them to UTF-8.

It would fix #417 

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -207,6 +207,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
+name = "chardetng"
+version = "0.1.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "14b8f0b65b7b08ae3c8187e8d77174de20cb6777864c6b832d8ad365999cf1ea"
+dependencies = [
+ "cfg-if",
+ "encoding_rs",
+ "memchr",
+]
+
+[[package]]
 name = "chrono"
 version = "0.4.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


yuma140902:
I think we need new tests.

yuma140902:
It seemed to be working fine, but I noticed that the `pueue log` command is not the only one that needs encoding conversion. 
In particular, it is difficult to reconcile `seek_to_last_lines()` with conversion.
I could create something like `DecodingReader` that implements `Read` and replace all `File`, but that is too big a job.
I'm beginning to feel that encoding conversion is outside the scope of this project.

Nukesor:
> I think we need new tests.

The whole integration test suite isn't running on windows yet, as nobody has tackled that ask yet.
This is definitely one of the bigger todos.

Nukesor:
I'll try to give it a closer look soon :) 
I'm currently quite busy at the moment. This probably needs a bit of focused time to deep-dive into this topic.

PR #419, opened 2023-02-21
Fix a few typos in CHANGELOG.md

AndreyAkinshin:


CHANGELOG.md
```diff
@@ -18,12 +18,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 
-- Allow to wait for specific task stati when using `pueue wait` [#400](https://github.com/Nukesor/pueue/issues/400).
+- Allow to wait for specific task state when using `pueue wait` [#400](https://github.com/Nukesor/pueue/issues/400).
```
Nukesor, line 21: Ok. this needs a bit help of a natural language speaker :D.
I'm still not 100% sure if I used `state` and `status` correctly in this project.

In my mind, tasks have a specific status at a given point in time. They're `Running`, `Done`, etc.
However, it's possible to query by or wait for a set of given "statuses"?
Now, "statuses" or "stati" don't seem to be valid words, even though at least "statuses" is theoretically (AFAIU) part of the official dictionary in the USA. But `status` is not supposed to be used in its plural form. One should apparently use the word `state` for this.

But in my mind, the a `state` of a task would be the whole state, including input, output, other metadata that's contained inside the daemon etc.

So here I am, confused about whether to use `statuses` or `states`, with lots of answers on the internet indicating to use `statuses`, even though it should (apparently) not be used :D

Nukesor: TIL, in German we actually use the latin declaration, which is `status` but with a long `u`.

AndreyAkinshin: @Nukesor thanks for providing a broad context for this naming problem!

So, which version do you prefer for the changelog? What should I do with this PR: close it or change the spelling?

Nukesor: Well, I was hoping you could shine a bit of light on that problem, as I'm really not sure which one is the correct translation :D

PR #420, opened 2023-02-25
fix: flush stdout

Banyc:
Thanks for sending a pull request!
Here's some info on how development works in this project:

The `main` branch is the version of the last release. \
`patch` level PRs, for example bugs or critical library updates, should be merged into `main` directly. \
A new `patch` release is usually published shortly after the PR is accepted.

New features should branch of the `development` branch. \

## Checklist

## Experiment

Previously, `stdout`s from between `pueue follow` and `pueue log` are different. `pueue log` prints a line more than `pueue follow` when the last line is still being produced.

I have made an experiment to proof the idea that it is `io::Stdout` not flushing the inner buffer:

```rust
#[cfg(test)]
mod tests {
    use std::{
        io::{self, Write},
        thread,
    };

    #[test]
    fn wo_flush() {
        let mut stdout = io::stdout();

        loop {
            stdout.write(b".").unwrap();

            thread::sleep(std::time::Duration::from_secs(1));
        }
    }

    #[test]
    fn with_flush() {
        let mut stdout = io::stdout();

        loop {
            stdout.write(b".").unwrap();
            stdout.flush().unwrap();

            thread::sleep(std::time::Duration::from_secs(1));
        }
    }
}
```

The `with_flush` test case prints `.` once every second but `wo_flush` does not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Fixed
+
+- Fixed changes to stdout not being printed after each I/O copy when using `pueue follow`.
+
 ## [3.1.2] - unreleased
 
 ## Fixed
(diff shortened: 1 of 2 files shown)


Banyc:
I have tested the newly built binaries. This change fixed the problem.

Nukesor:
Thanks for the fix :)

The remote streaming logic already flushes `stdout`, so it only makes sense to flush it on local follows as well.

PR #458, opened 2023-08-18
add: make shell command configurable

### LINKED ISSUE #363: Make shell configurable
Hi, thank you so much for your great work, after using it for a while I found that I fell in love with it :-D I'm using it to schedule my background tasks.

Currently I'm trying to make [nushell](https://www.nushell.sh/) supports background task, and I want to implemented it based on `pueued` and `pueue-lib` :)

After investigating code, I found that pueued runs `sh -c xxx` on linux/mac, and runs `powershell -c xxx` on windows.  It would be awesome to support a configuration point where other shells, like nushell, could be configured to be used instead of only bash and powershell. For instance, I'd like to call pueue with `nu -c xxx`, but I cannot right now.

Make internal shell configurable while running pueue and pueued.  An example would be something like this:
```yaml
shared:
  shell: nu
```

If the `shell` configuration doesn't exists, make it as default like currently.

None

#### Additional context

I have changed one line by myself for experimental:
```rust
pub fn compile_shell_command(command_string: &str) -> Command {
    let mut command = Command::new("nu");  // changed line
    command.arg("-c").arg(command_string);

    command
}
```

And running pueue, I get the following result:
Screen Shot 2022-09-16 at 22 40 40

Additionally, if we support this, we can even support adding running powershell commands on linux/mac, because powershell supports many platform for now (https://github.com/PowerShell/PowerShell#get-powershell)

One thing to consider is environment transfer to `pueued`, I'm considering build a client based on `pueue-lib`(because for querying job status, I need to generate a datastructure that nushell used, it can be implemented there)

### LINKED ISSUE #454: Expand aliases

Currently, the following will fail:
```
alias foo=date
pueue add foo
```

It fails because in the `pueue` context, user aliases are not loaded, so `foo` is unknown.
It would be useful to have an option (set as default ?) that loads user's env such as currently set aliases.
Adding `shopt -s expand_aliases` may help but I did not test.

`pueue` is very useful for a user to queue up a bunch of tasks. It is quick natural for users to already have created aliases for the individual tasks. For instance, a user may already have an alias to run `freshlam && scan` or `brew update && brew upgrade`.

### Alternatives

_No response_

### Additional context

_No response_

Nukesor:
Fixes #454
Fixes #363 

This is waiting for some people to actually test this.

I would like to see some tests for:

- [x] Bash
- [x] Zsh ( if it's possible)
- [ ] Oilshell
- [ ] Fish
- [x] Nushell 
- [x] Powershell

However, the more the merrier.

How to test:

```sh
cd /tmp
git clone git@github.com:nukesor/pueue
cd pueue
git switch configurable-shell
cargo install --path pueue
```

For bash, add this to your daemon settings section:
```yml
  shell_command: [
    "bash",
    "-c",
    "shopt -s expand_aliases && {{ pueue_command_string }}",
  ]
  env_vars:
    BASH_ENV: "/path/to/aliases_file"
```

For zsh, add this:
```yml
  shell_command: [
    "zsh",
    "-c",
    ". /path/to/aliases_file; setopt aliases; eval {{ pueue_command_string }}",
  ]
```

Make sure to restart your daemon!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - QoL improvement: Don't pause groups if there're no queued tasks. [#452](https://github.com/Nukesor/pueue/issues/452)
   Auto-pausing of groups was only done to prevent the unwanted execution of other tasks, but this isn't necessary, if there're no queued tasks.
 
+### Added
+
+- Allow configuration of the shell command that executes task commands. [#454](https://github.com/Nukesor/pueue/issues/454)
+
 ## [3.2.0] - 2023-06-13
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Added
 
 - Allow configuration of the shell command that executes task commands. [#454](https://github.com/Nukesor/pueue/issues/454)
+- Allow injection of hard coded environment variables via config file. [#454](https://github.com/Nukesor/pueue/issues/454)
 
 ## [3.2.0] - 2023-06-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 
-- Allow configuration of the shell command that executes task commands. [#454](https://github.com/Nukesor/pueue/issues/454)
-- Allow injection of hard coded environment variables via config file. [#454](https://github.com/Nukesor/pueue/issues/454)
+The two following features are very new and marked as "experimental" for the time being.
+They might be reworked in a later release, since working with shells is always tricky and this definitely need more testing.
+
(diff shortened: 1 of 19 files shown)


Isti115:
Thanks for taking the effort to support multiple shells!
I have managed to get my `nushell` definitions working with this configuration:
```
  shell_command: [
    "nu",
    "--config .config/nushell/config.nu",
    "--env-config .config/nushell/env.nu",
    "-c {{ pueue_command_string }}"
  ]
```

WindSoilder:
Sorry for late reply, I've tried in nushell recently, and verified it works too

Nukesor:
I think I'm going to merge this, It has been open for long enough and from what I can see it works quite well with various shells :) 

We'll squash any upcoming bugs once the feature is released.

I'll explicitly mark this feature as experimental, since there's a lot of stuff that can go wrong with shells :D

jarrodu:
Is this feature documented? I did not see it in the the Wiki. I found the config option in the config file and searched for "shell_command" here.

Nukesor:
It's not yet documented, we should add a section to the wiki :sweat_smile: eventually.

The feature was only announced in the recent changelog of https://github.com/Nukesor/pueue/releases/tag/v3.3.0
Most info about this can be found in this issue: https://github.com/Nukesor/pueue/issues/454

v217:
Hi,
Thank you for pueue!
For the fish shell, this works for me:

```yaml
shell_command: [
    "fish",
    "-c",
    "{{ pueue_command_string }}"
  ]
```

PR #504, opened 2024-03-04
fix: make graceful daemon shutdown exit with code 0

rycee:
## Description

When shutting down pueued using SIGINT or SIGTERM then the exit code should be zero. This is necessary, for example, when running as a systemd service since systemd will otherwise put the service in a failed state whenever it shuts down.

I'm not sure why it is currently exiting with an error-code. Maybe it is for some good reason? But looking at the code comments it definitely mentions exiting _gracefully_, which I would interpret as exit code 0.

## Checklist

- [X] I _think_ I picked the correct source and target branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Fix reading of configuration files that lacks a `shared` section.
 
+- Made the daemon exit gracefully (exit code 0) on SIGINT and SIGTEM.
+
 ## [3.3.3] - 2024-01-04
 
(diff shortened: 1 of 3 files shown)


Nukesor:
Good point. I'm no longer sure why I let it exit with code 1 to be honest. But I tested for it xD!

Exiting with `0` seems like the correct thing to do :+1:

rycee:
Cool, thanks! I've rebased the branch with the conflict resolved.

rycee:
Thanks!

Nukesor:
Thank you! I'll have to release a new version soon with these changes :D

rycee:
That would be great, much appreciated!

PR #505, opened 2024-03-04
fix: handle settings with missing `shared` section

rycee:
## Description

This allows configuration files without a `shared` field, using the default value if missing.

Before:

    $ pueued -vv -c /dev/null
    Error: Error while reading configuration.

    Caused by:
        Error while reading configuration:
        missing field `shared`

After:

    $ pueued -vv -c /dev/null
    21:12:26 [INFO] Restoring state
    21:12:26 [INFO] Using unix socket at: "/run/user/1000/pueue_rycee.socket"

## Checklist

- [x] I _think_ I picked the correct source and target branch.


Nukesor:
Thanks!

PR #512, opened 2024-03-18
issue#508 rudimentary bug fix

Shakil582:
Thanks for sending a pull request!
Here's some info on how development works in this project:

The `main` branch is the version of the last release. \
`patch` level PRs, for example bugs or critical library updates, should be merged into `main` directly. \
A new `patch` release is usually published shortly after the PR is accepted.

New features should branch of the `development` branch. \

## Checklist

- [* ] I picked the correct source and target branch.
- [* ] I included a new entry to the `CHANGELOG.md`.
- [ *] I checked `cargo clippy` and `cargo fmt`. The CI will fail otherwise anyway.
- [ *] (If applicable) I added tests for this feature or adjusted existing tests.
- [ *] (If applicable) I checked if anything in the wiki needs to be changed.

I've never made a pull request for a github repo, this is my first open source contribution. 
That being said what's happening here is that we are using first as the limit but we don't remove the unused groups from the list of tasks so we are grabbing the first task which is from g1 but it won't get printed when it tries to print only the used groups.

### DIFF
--- a/pueue/src/client/display/state.rs
+++ b/pueue/src/client/display/state.rs
@@ -34,6 +34,9 @@ pub fn print_state(
     if let Some(query) = query {
         let query_result = apply_query(&query.join(" "))?;
         table_builder.set_visibility_by_rules(&query_result.selected_columns);
+        if let Some(group) = &group_only {
+            tasks = query_result.apply_groups(tasks, group);
+        }
         tasks = query_result.apply_filters(tasks);
         tasks = query_result.order_tasks(tasks);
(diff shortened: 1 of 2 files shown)


Shakil582:
Just realized I should remove the print_single_group and rename print_all_groups and just call that since the groups are being filtered out now. Will do it later today.

Nukesor:
This logic only fixes one of the symptoms, but not the original problem :)

I created a fix for the actual problem over here: https://github.com/Nukesor/pueue/pull/513

Could you try and see if this fixes your problem :)?

PR #527, opened 2024-04-23
Feature: add shell completions for `nushell`

Tirka:
## Rationale
pueue recommended as a ~~temporary workaround~~ solution for background task spawning for nushell users.

It would be nice to have shell completions within nushell.

## Usage
Currently nushell has no preferred way to install completion files into system: https://github.com/nushell/nushell/issues/11337

One possible way of doing this could be:

1) generate completions

```nushell
mkdir /usr/share/nushell-completions/pueue/
pueue completions nushell /usr/share/nushell-completions/pueue/
mv /usr/share/nushell-completions/pueue/pueue.nu /usr/share/nushell-completions/pueue/mod.nu
```

2) open nushell config

```nushell
config nu
```

3) add reference to completion file into config (at the bottom)

```text
use ~/.config/nushell/completions/pueue/ *
```

4) save config and restart the shell

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Nushell autocompletion script [#527](https://github.com/Nukesor/pueue/pull/527)
+
 ## [3.4.0] - 2024-03-22
 
 ### Added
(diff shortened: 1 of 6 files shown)


Tirka:
BTW

Not sure why this code block repeats twice, looks like a bug

https://github.com/Nukesor/pueue/blob/cc7aedd9889e3a24337e4bc2eb50b1722267a32c/pueue/src/bin/pueue.rs#L104-L144

Nukesor:
Yep, not sure how that managed to sneak in there. Feel free to delete that duplicate blob. Probably a copy past error by me when i introduced the feature to print completions to stdout.

Also, could you add nushell to the completions integrationt test :)?

Tirka:
> Also, could you add nushell to the completions integrationt test :)?

The existing test has already been corrected. Should I add something else?

Nukesor:
Nice, thanks for the contribution :)
Change looks good to me!

PR #531, opened 2024-05-13
Add FreeBSD process helper to facilitate FreeBSD builds

zi0r:
These patches are already part of the FreeBSD port for pueue.  Including them here will make port updates a little easier.

You can view existing applied patches [here](https://cgit.freebsd.org/ports/tree/deskutils/pueue/files).

To completely fix the build, we also need to remove procfs/libproc from the Cargo files [while building under Freebsd].  I wasn't sure on the best way to achieve this, so I am simply noting it here.

## Checklist

- [X] I included a new entry to the `CHANGELOG.md`.
- [X] I checked `cargo clippy` and `cargo fmt`. The CI will fail otherwise anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Added
 
 - Nushell autocompletion script [#527](https://github.com/Nukesor/pueue/pull/527)
+- Add FreeBSD process helper to facilitate FreeBSD builds
 
 ## \[3.4.0\] - 2024-03-22
 
--- a/pueue/Cargo.toml
+++ b/pueue/Cargo.toml
@@ -65,5 +65,7 @@ crossterm = { version = "0.27", default-features = false, features = ["windows"]
 # Test specific dev-dependencies
 [target.'cfg(any(target_os = "linux", target_os = "freebsd"))'.dependencies]
 whoami = "1"
-procfs = { version = "0.16", default-features = false }
 
+# Test specific Linux dev-dependencies
+[target.'cfg(target_os = "linux")'.dependencies]
+procfs = { version = "0.16", default-features = false }
--- a/pueue_lib/Cargo.toml
+++ b/pueue_lib/Cargo.toml
@@ -67,8 +67,10 @@ winapi = { version = "0.3", features = [
 
 # Unix
 [target.'cfg(unix)'.dependencies]
-libproc = "0.14.6"
 whoami = "1"
+
+[target.'cfg(any(target_os = "linux", target_os = "macos"))'.dependencies]
+libproc = "0.14.6"
 
 # Linux only
 [target.'cfg(target_os = "linux")'.dependencies]
--- a/pueue_lib/src/process_helper/freebsd.rs
+++ b/pueue_lib/src/process_helper/freebsd.rs
@@ -1,0 +1,15 @@
+use std::path::Path;
+
+/// Check, whether a specific process is exists or not
+pub fn process_exists(pid: u32) -> bool {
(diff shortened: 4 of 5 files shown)


Nukesor:
Check the `pueue/Cargo.toml` at the very bottom.

Change this:
```toml
# Test specific dev-dependencies
[target.'cfg(any(target_os = "linux", target_os = "freebsd"))'.dependencies]
whoami = "1"
procfs = { version = "0.16", default-features = false }
```

To this:
```toml
# Test specific Linux + BSD specific dev-dependencies
[target.'cfg(any(target_os = "linux", target_os = "freebsd"))'.dependencies]
whoami = "1"

# Test specific Linux dev-dependencies
[target.'cfg(target_os = "linux")'.dependencies]
procfs = { version = "0.16", default-features = false }
```


In `pueue_lib/Cargo.toml`, the following needs to be changed:
```toml
# Unix
[target.'cfg(unix)'.dependencies]
libproc = "0.14.6"
whoami = "1"

# Linux only
[target.'cfg(target_os = "linux")'.dependencies]
procfs = { version = "0.16", default-features = false }
```
Becomes
```toml
# Unix
[target.'cfg(unix)'.dependencies]
whoami = "1"

[target.'cfg(any(target_os = "linux", target_os = "macos"))'.dependencies]
libproc = "0.14.6"

# Linux only
[target.'cfg(target_os = "linux")'.dependencies]
procfs = { version = "0.16", default-features = false }
```


So. If Pueue is to have proper support for FreeBSD, I want this to be properly tested.

Please:
- Add tests
- Make sure the test suite runs on FreeBsd.
  -> Unit tests should succeed, integration tests should work if possible, otherwise just don't build them as can be seen in `pueue/tests/tests.rs`. From what I understand, the test worked at some point for FreeBSD and actually used `procfs`, which is why I'm a bit confused that it's now no longer allowed.
- Investigate how to run the test suite for FreeBSD in the CI pipeline. The best way to achieve this is to fork the repo and test the pipelines on your own repo.

zi0r:
Thanks for the tip!  I've added an additional commit to break out the procfs support further.  While FreeBSD does still have a way to mount procfs, it isn't enabled by default.  This adds another step for users and is wonderful to be able to bypass it whenever we have another means.

Unfortunately, I don't believe I have time to dedicate to revamping the tests suite or CI Pipeline for FreeBSD support.  However, I can tell you that the commits in this PR (or a close variation of them) have been in use in the FreeBSD ports tree since pueue v3.0.0, when it was first added on Dec 25 2022.  You can view the history [here](https://www.freshports.org/deskutils/pueue).

Lastly, there is one additional patch that we include (adopted from alpine linux):
- patch from https://git.alpinelinux.org/aports/commit/community/pueue?id=6e6e3b80434f09b1ec0e777afa2c88ab526b0ba3
- bug#272467
- https://cgit.freebsd.org/ports/tree/deskutils/pueue/files/patch-fix-build?id=f10f5b39edb7a37a61d7bd07baeba1da0a33c859

I'm happy to add this to my PR, if desired.

Nukesor:
It's super fascinating to see how pueue is packaged :D  
I'm ususally not involved into any packaging work at all, but there's quite a lot of stuff going on over there.

In the future, I would be more than happy to release a new patch whenever something breaks due to changes in the build environment/toolchain. Just give me a ping :)

zi0r:
> It's super fascinating to see how pueue is packaged :D I'm ususally not involved into any packaging work at all, but there's quite a lot of stuff going on over there.
> 
> In the future, I would be more than happy to release a new patch whenever something breaks due to changes in the build environment/toolchain. Just give me a ping :)

This sounds amazing and will certainly make the updating process on the FreeBSD ports tree side of things much easier!  Thanks!

Nukesor:
Fixed that lint for you :) 

Thanks for the contribution! I'll release a patch soon.

PR #538, opened 2024-06-04
fix: Releases pipeline

Nukesor:
Release on ubuntu breaks currently and the CI is way to too complicated.


Nukesor:
Works just fine, tested over [here](https://github.com/Nukesor/testing/actions/runs/9375337782) in my testing repo.

PR #544, opened 2024-06-15
feat: add configurable socket permissions

JP-Ellis:
## Summary

Have `pueue` set the permissions of the socket created. Previously, the permissions where unspecified and (at least for me) defaulted to `755`. The permissions can be configured through the new `unix_socket_permissions` setting.

As part of this change, the default settings are moving to `700` which is more restricted than the (possibly system-dependent) previous default. The more restricted permissions should not have any practical impact as users without write permissions would not be able to use the socket, though removing read access to others may possibly improve security.

## Checklist

### DIFF
--- a/pueue_lib/src/network/socket/unix.rs
+++ b/pueue_lib/src/network/socket/unix.rs
@@ -1,10 +1,12 @@
 use std::convert::TryFrom;
+use std::fs::{set_permissions, Permissions};
+use std::os::unix::fs::PermissionsExt;
 
 use async_trait::async_trait;
 use log::info;
 use rustls::pki_types::ServerName;
 use tokio::io::{AsyncRead, AsyncWrite};
-use tokio::net::{TcpListener, TcpStream, UnixListener, UnixStream};
+use tokio::net::{TcpListener, TcpStream, UnixListener, UnixSocket, UnixStream};
 use tokio_rustls::TlsAcceptor;
 
 use crate::error::Error;
@@ -142,8 +144,29 @@ pub async fn get_listener(settings: &Shared) -> Result<GenericListener, Error> {
             })?;
         }
 
-        let unix_listener = UnixListener::bind(&socket_path)
-            .map_err(|err| Error::IoPathError(socket_path, "creating unix socket", err))?;
+        // The various nix platforms handle socket permissions in different
+        // ways, but generally prevent the socket's permissions from being
+        // changed once it is being listened on.
+        let socket = UnixSocket::new_stream()
+            .map_err(|err| Error::IoError("creating unix socket".to_string(), err))?;
+        socket.bind(&socket_path).map_err(|err| {
+            Error::IoPathError(socket_path.clone(), "binding unix socket to path", err)
+        })?;
+
+        if let Some(mode) = settings.unix_socket_permissions {
+            set_permissions(&socket_path, Permissions::from_mode(mode)).map_err(|err| {
+                Error::IoPathError(
+                    socket_path.clone(),
+                    "setting permissions on unix socket",
+                    err,
+                )
+            })?;
+        }
+
+        let unix_listener = socket.listen(1024).map_err(|err| {
+            Error::IoPathError(socket_path.clone(), "listening on unix socket", err)
+        })?;
+
         return Ok(Box::new(unix_listener));
     }
 
--- a/pueue_lib/src/settings.rs
+++ b/pueue_lib/src/settings.rs
@@ -46,6 +46,12 @@ pub struct Shared {
     /// The path to the unix socket.
     #[cfg(not(target_os = "windows"))]
     pub unix_socket_path: Option<PathBuf>,
+    /// Unix socket permissions. Typically specified as an octal number and
+    /// defaults to `0o700` which grants only the current user access to the
+    /// socket. For a client to connect to the daemon, the client must have
+    /// read/write permissions.
+    #[cfg(not(target_os = "windows"))]
+    pub unix_socket_permissions: Option<u32>,
 
     /// The TCP hostname/ip address.
     #[serde(default = "default_host")]
@@ -147,6 +153,8 @@ impl Default for Shared {
             unix_socket_path: None,
             #[cfg(not(target_os = "windows"))]
             use_unix_socket: true,
+            #[cfg(not(target_os = "windows"))]
+            unix_socket_permissions: Some(0o700),
             host: default_host(),
             port: default_port(),
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ The new state design fixes this issue, which allows Pueue to do subprocess state
 - Add `--all` and `--group` to `pueue log`. [#509](https://github.com/Nukesor/pueue/issues/509)
 - Add `pueue reset --groups [group_names]` to allow resetting individual groups. [#482](https://github.com/Nukesor/pueue/issues/482) \
   This also refactors the way resets are done internally, resulting in a cleaner code architecture.
+- Ability to set the Unix socket permissions through the new `unix_socket_permissions` configuration option. [#544](https://github.com/Nukesor/pueue/pull/544)
 
 ## \[3.4.1\] - 2024-06-04
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -858,7 +858,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0c2a198fb6b0eada2a8df47933734e6d35d350665a33a3593d7164fa52c75c19"
 dependencies = [
  "cfg-if",
- "windows-targets 0.48.5",
+ "windows-targets 0.52.5",
 ]
 
 [[package]]
--- a/pueue/tests/daemon/integration/mod.rs
+++ b/pueue/tests/daemon/integration/mod.rs
@@ -16,6 +16,7 @@ mod restart;
 mod restore;
 /// Tests for shutting down the daemon.
 mod shutdown;
+mod socket_permissions;
 mod spawn;
 mod start;
 mod stashed;
--- a/pueue/tests/daemon/integration/socket_permissions.rs
+++ b/pueue/tests/daemon/integration/socket_permissions.rs
@@ -1,0 +1,56 @@
+use anyhow::Result;
+use std::fs;
+use std::os::unix::fs::PermissionsExt;
+
+use anyhow::Context;
+
+use crate::helper::*;
+
+/// Make sure that the socket permissions are appropriately set.
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+#[cfg(not(target_os = "windows"))]
+async fn test_socket_permissions_default() -> Result<()> {
+    let (settings, _tempdir) = daemon_base_setup()?;
+    let shared = &settings.shared;
+    let mut child = standalone_daemon(shared).await?;
+
+    assert_eq!(
+        fs::metadata(shared.unix_socket_path())?
+            .permissions()
+            .mode()
+            // The permissions are masked with 0o777 to only get the last 3
+            // digits.
+            & 0o777,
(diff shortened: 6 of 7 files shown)


JP-Ellis:
A couple of things outstanding for this PR before merging:

- I have not added a test yet. You have quite an exhaustive test suite, so I would like you to point me in the right direction as to what tests to add.
- There is a test failing when it comes to the backwards compatibility. It wasn't immediately clear to me how this should be resolved.

pueue_lib/src/network/socket/unix.rs
```diff
@@ -143,7 +145,14 @@ pub async fn get_listener(settings: &Shared) -> Result<GenericListener, Error> {
...
 
         let unix_listener = UnixListener::bind(&socket_path)
-            .map_err(|err| Error::IoPathError(socket_path, "creating unix socket", err))?;
+            .map_err(|err| Error::IoPathError(socket_path.clone(), "creating unix socket", err))?;
+
+        // Adjust socket permissions
+        let mut permissions = fs::metadata(&socket_path)
+            .map_err(|err| Error::IoPathError(socket_path, "obtaining file metadata", err))?
+            .permissions();
+        permissions.set_mode(settings.unix_socket_permissions);
```
Nukesor, line 154: That config is an option, hide this behind a `if let Some()` :)

Nukesor (changes requested):
Regarding tests, the way to do it would be to start a daemon with settings that sets specific socket permissions.

A good test that can be used for some guidance is `tests/daemon/integration/environment_variables.rs`. 

I think it would be nice to have a new `unix_socket.rs` test file.

Nukesor:
I have a few waiting functions in `tests/helpers/wait.rs`.
Just take one of those and modify them to wait for a file to show up :)

I also edited my original comment to contain a bit more info, but I noticed you already read it. You're too quick :D

Nukesor:
Also, a new lint popped up due to Rust 1.79.
It's already fixed on main :)

Nukesor:
Do you need any further help :)?

JP-Ellis:
Just need a bit more time 😆 I've had a busy few days lately, but I should have time by the end of the week to finish it up 👍

Nukesor:
Don't worry, take your time. Just wanted to check in if you need anything :)

JP-Ellis:
So I finally got back to this PR, and I have some not-so-great news. 

Unfortunately the [`PermissionsExt`](https://doc.rust-lang.org/stable/std/os/unix/fs/trait.PermissionsExt.html) from the standard library does not work as expected, with `set_mode` have no effect on the socket. This incompatibility [has been reported by others](https://users.rust-lang.org/t/creating-a-unix-domain-socket-with-a-particular-mode/72940). 

The solution in their case is to use the unsafe `libc::umask` method which is unfortunate for two reasons:

1. It introduces a dependency on `libc` (though it appears `pueue` already dependency on `libc` transitively).
2. It introduces `unsafe` code.


I have added a separate commit implementing this, and while it works well for the hardcoded default value, I was unable to get the test with the modified value to work. For some reason, it always defaults back to the value in `pueue_lib/src/settings.rs` irrespective of being modified in the test. I might be missing something simple?

pueue_lib/src/network/socket/unix.rs
```diff
@@ -142,8 +142,19 @@ pub async fn get_listener(settings: &Shared) -> Result<GenericListener, Error> {
...
+
         let unix_listener = UnixListener::bind(&socket_path)
-            .map_err(|err| Error::IoPathError(socket_path, "creating unix socket", err))?;
+            .map_err(|err| Error::IoPathError(socket_path.clone(), "creating unix socket", err))?;
+
+        // Reset the umask now that the socket is created.
+        if let Some(permissions) = old_umask {
+            unsafe { libc::umask(permissions) };
+        }
+
```
Nukesor, line 157: In the case of an error, the umask would be still set to the wrong value.

We cannot use the `?` operator directly, but rather use it after the reset of the umask below.

pueue_lib/src/network/socket/unix.rs
```diff
@@ -142,8 +142,19 @@ pub async fn get_listener(settings: &Shared) -> Result<GenericListener, Error> {
...
+            .unix_socket_permissions
+            .map(|permissions| unsafe { libc::umask(!permissions) });
+
         let unix_listener = UnixListener::bind(&socket_path)
-            .map_err(|err| Error::IoPathError(socket_path, "creating unix socket", err))?;
+            .map_err(|err| Error::IoPathError(socket_path.clone(), "creating unix socket", err))?;
+
+        // Reset the umask now that the socket is created.
+        if let Some(permissions) = old_umask {
+            unsafe { libc::umask(permissions) };
```
Nukesor, line 155: Another problem with this approach is that Pueue is multi-threaded.

From what I understand, setting the `umask` affects the whole process.
And the umask is also inherited to child processes.

-> We have to make really sure that 
- No files are written while this happens
- No subprocesses of any kind are started while this happens.

Nukesor (changes requested):
After thinking about the possible side-effects of using `umask`, I'm not convinced that this is the solution we should use, especially since we're dealing with a multi-threaded environment that spawns subprocesses and reads/writes files.

I just played around with the unix socket on my local PC and I could easily change the permissions of the socket via `chmod 777`, so there's theoretically no need to set the umask directly.

There's obviously that bug in the stdlib, but isn't there another way to set file permissions?

The nix crate uses `fchmod`:
https://docs.rs/nix/latest/src/nix/sys/stat.rs.html#272-276

JP-Ellis:
> After thinking about the possible side-effects of using `umask`, I'm not convinced that this is the solution we should use, especially since we're dealing with a multi-threaded environment that spawns subprocesses and reads/writes files.

Completely agree with your assessment. I wasn't too happy with using `umask` myself and the need to set/reset this while creating the socket. I still wanted to test it and share my findings :)


> There's obviously that bug in the stdlib, but isn't there another way to set file permissions?
> 
> For example, the nix crate uses `fchmod`: https://docs.rs/nix/latest/src/nix/sys/stat.rs.html#272-276

More than happy to use other crates, I just wanted to avoid unnecessarily introducing new dependencies (unless they exist transitively).

I've updated the PR to make use of the `nix` crate.

Nukesor:
This thread is pretty informative:
https://stackoverflow.com/a/74329441

It would explain why stdlib didn't work as expected.

Still, I'm curious why I'm able to change permissions on the unix file socket from the outside.
From what I understand, `chmod 777 $XDG_RUNTIME_DIR/pueue_$USER.socket` should be a no-op/throw an exception. But it works just fine.

Is that only forbidden for the currently attached process? I'm a bit confused.
Which Operating System are you on right now? The behavior seems to differ in different environments. I'm on Arch Linux right now.

Nukesor:
So, Tokio uses `mio` [under the hood](https://github.com/tokio-rs/tokio/blob/master/tokio/src/net/unix/listener.rs#L73) for their unix socket handling.

Mio's in turn has a [bind_addr](https://github.com/tokio-rs/mio/blob/master/src/sys/unix/uds/listener.rs#L12) function which uses the `new_socket` function.

[new_socket](https://github.com/tokio-rs/mio/blob/master/src/sys/unix/net.rs#L15) sadly always calls the `socket` syscall, which as far as I understood, always tries to create a new socket.

If it indeed always tries to create a new function, we might need to patch `mio` to allow using an existing socket (if it exists) with a fallback to creating a socket when it doesn't.
`mio` should then expose their `new_socket` function, so we can use it to create our own socket before binding to it.


I'll take a closer look in the next few days :)

Nukesor:
Ok, in theory, we can create our own unix socket, set its permission and attach a UnixListener to it.

Tokio allows the creation of a UnixListener from a stdlib UnixListener [over here](https://docs.rs/tokio/latest/tokio/net/struct.UnixListener.html#method.from_std).

And a stdlib UnixListener can [be created via a socket_addr path](https://doc.rust-lang.org/nightly/std/os/unix/net/struct.UnixListener.html#method.bind_addr).

We just have to find out how to create raw unix sockets.

I'll be able to take a closer look at this at the end of next week, I'm pretty busy right now :sweat_smile:

JP-Ellis:
Wow! That's some good digging!

I certainly could implement the logic you're talking about whereby the socket is created first, then `chmod`, and finally bound with tokio. I honestly would not have thought of that first. 

I'll look at raising a ticket to request setting permissions with the upstream libraries. It seems like a bit of a gap given what you have found. 

Lastly, I've been doing this testing in macOS, though I also run Arch Linux (and would need this feature on my small Arch 'server').

Thanks so much for the help though!

JP-Ellis:
Finally got it to work as intended!

Ultimately needed to use tokio's `UnixSocket` to create the socket, and then `fchmodat` to set the permissions on the path _before_ listening on the socket.

From a pure security perspective, this implementation does have a very minor vulnerability in the time between the binding of the socket to a path, and the subsequent setting of the permissions. I doubt this will be an issue, but I wanted to flag that just in case. I did try setting the permissions on the socket's underlying `RawFd` _before_ binding it to a path, but this did not work as intended[^1]

[^1]: On macOS, the `chmod` call resulted in an error; and on the Ubuntu runners, the `chmod` call resulted in a silent no-op.

pueue_lib/src/network/socket/unix.rs
```diff
@@ -142,8 +142,35 @@ pub async fn get_listener(settings: &Shared) -> Result<GenericListener, Error> {
...
+                nix::sys::stat::Mode::from_bits_truncate(mode),
+                nix::sys::stat::FchmodatFlags::FollowSymlink,
+            )
+            .map_err(|err| {
+                Error::IoPathError(
+                    socket_path.clone(),
+                    "setting socket permissions",
+                    std::io::Error::from_raw_os_error(err as i32),
+                )
+            })?;
```
Nukesor, line 167: I figured out how to do this without nix:

```
        if let Some(mode) = settings.unix_socket_permissions {
            set_permissions(&socket_path, Permissions::from_mode(mode)).map_err(|err| {
                Error::IoPathError(
                    socket_path.clone(),
                    "setting permissions on unix socket",
                    err,
                )
            })?;
        }
```

This uses just the std lib and works at least for me :)
Could you test whether this also works for your environment?

If so, please remove the nix dependency :)

pueue_lib/src/settings.rs
```diff
@@ -46,6 +46,12 @@ pub struct Shared {
     /// The path to the unix socket.
     #[cfg(not(target_os = "windows"))]
     pub unix_socket_path: Option<PathBuf>,
+    /// Unix socket permissions. Typically specified as an octal number and
+    /// defaults to `0o700` which grants only the current user access to the
+    /// socket. For a client to connect to the daemon, the client must have
+    /// read/write permissions.
+    #[cfg(not(target_os = "windows"))]
+    pub unix_socket_permissions: Option<nix::sys::stat::mode_t>,
```
Nukesor, line 54: Change to `u32`. `mode_t` is just a type alias:
https://docs.rs/nix/latest/nix/sys/stat/type.mode_t.html

Nukesor (changes requested):

pueue_lib/src/network/socket/unix.rs
JP-Ellis: Sure, let me see if it works for me.

pueue_lib/src/settings.rs
JP-Ellis: If the `nix` dependency is removed, then this point is moot. 

Otherwise, the `mode_t` is still required because the `mode_t` is a different alias on different machines. On macOS for example, `mode_t` is an alias for `u16`.

pueue_lib/src/network/socket/unix.rs
JP-Ellis: Can confirm that also works, so I've removed the `nix` dependency 🥳  PR is updated

Nukesor (approved):

Issue #554, opened 2024-07-20
Allow `pueue status` to order tasks by `enqueue_at`


pacien:
`pueue status order_by enqueue_at [asc/desc]`
does not appear to order tasks as requested.

Minimal non-working example:

```fish
❯ pueue add --delay 2024-07-21 echo hi
New task added (id 0). It will be enqueued at 2024-07-21 00:00:00

❯ pueue add --delay 2024-07-22 echo hi
New task added (id 1). It will be enqueued at 2024-07-22 00:00:00

❯ pueue add --delay 2024-07-21 echo hi
New task added (id 2). It will be enqueued at 2024-07-21 00:00:00

❯ pueue status order_by enqueue_at
Group "default" (1 parallel): running
───────────────────────────────────────────────────────────────
 Id   Status    Enqueue At   Command   Path        Start   End 
═══════════════════════════════════════════════════════════════
 2    Stashed   00:00:00     echo hi   /home/user               
───────────────────────────────────────────────────────────────
 1    Stashed   2024-07-22   echo hi   /home/user               
                00:00:00                                       
───────────────────────────────────────────────────────────────
 0    Stashed   00:00:00     echo hi   /home/user               
───────────────────────────────────────────────────────────────

❯ pueue status order_by enqueue_at desc
Group "default" (1 parallel): running
───────────────────────────────────────────────────────────────
 Id   Status    Enqueue At   Command   Path        Start   End 
═══════════════════════════════════════════════════════════════
 0    Stashed   00:00:00     echo hi   /home/user               
───────────────────────────────────────────────────────────────
 1    Stashed   2024-07-22   echo hi   /home/user               
                00:00:00                                       
───────────────────────────────────────────────────────────────
 2    Stashed   00:00:00     echo hi   /home/user               
───────────────────────────────────────────────────────────────
```

Expected output:
task ID 1 shouldn't be in-between in either cases.

Platform info:
- pueue 3.4.0
- NixOS 24.05 (GNU/Linux 6.6.32, x86_64)

Nukesor:
Feel free to implement this :)

Nukesor:
You're right, I completely forgot about those docs as I didn't write them myself. Sorry for that.

Yep, according to the docs, `enqueue_at` should be possible to be ordered by. Should be fairly straight forward to add it to https://github.com/Nukesor/pueue/blob/main/pueue/src/client/query/mod.rs#L65

[...]

PR #559, opened 2024-07-26
add: allow status order_by enqueue_at

pacien:
GitHub: closes https://github.com/Nukesor/pueue/issues/554

## Checklist

### DIFF
--- a/pueue/src/client/query/mod.rs
+++ b/pueue/src/client/query/mod.rs
@@ -2,6 +2,7 @@
 #![allow(clippy::empty_docs)]
 
 use anyhow::{bail, Context, Result};
+use chrono::prelude::*;
 use pest::Parser;
 use pest_derive::Parser;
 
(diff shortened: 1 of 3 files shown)


pueue/src/client/query/mod.rs
```diff
@@ -93,6 +94,24 @@ impl QueryResult {
...
+                            enqueue_at: Some(enqueued_at),
+                            ..
+                        } => *enqueued_at,
+                        // considered far in the future when no explicit date:
+                        _ => DateTime::<Utc>::MAX_UTC.into(),
+                    }
+                }
+
+                enqueue_date(task1).cmp(&enqueue_date(task2))
+            }
```
Nukesor, line 114: `enqueue_at` and `enqueued_at` are two different things though.

One is when it has been enqueued, the other one when it'll be enqueued.

Nukesor (changes requested):

pueue/src/client/query/mod.rs
Nukesor: True, but the column that's shown in the `pueue status` column is the `enqueue_at` field. I'm a bit concerned that it might be non-intuitive for users on how those commands are sorted, since they don't see the `enqueued_at` field.

Not showing `enqueued_at` in `pueue status` output was a deliberate decision as this has never been asked for and the `pueue status` table should be as narrow as possible to make it somewhat

Though this isn't a strong feeling on my end. So if you think that this feels intuitive, I would be ok with the current implementation :)

Nukesor: Fair point, the fallback behavior would be just as confusing.

You convinced me, let's keep it the way it is right now :)

Nukesor:
Thanks for your contribution :)

PR #595, opened 2025-02-01
Immediate follow

### LINKED ISSUE #592: feature suggestion: --folow argument for `pueue add`

I would like to have a simple `--follow` flag to `add`. Running `pueue add --folow <CMD>` should act just like:
```bash
pueue add <CMD>
pueue follow <TASK_ID_FROM_ADD>
```
Basically, it adds the task and follows its output. Just like the separate commands, Ctrl+C during a `pueue add --folow` should stop following but continue running the task.

I am not quite sure what the behaviour should be if the task cannot be immediately started. Running separate add/follow will just block while waiting for the command to start. Maybe it should print a warning then continue waiting.

This will help get instant feedback when there is an initialisation problem for long-running tasks. A lot of the time, I run `pueue add` but make simple mistakes in the command (e.g. typo, wrong working directory). The mistakes cause the task to fail immediately but this is not immediately visible.

At the moment, I have to remember to `pueue status` after every `add` to make sure that the command is running properly. It's easy to forget this step, go away, and come back to an error instead of the the results I waited for.

### Alternatives

One could make an alias, however this makes it harder to add more options to `add`. I guess this bash would work:
```bash
pueue-add-follow() {
  task=$(pueue add --print-task-id "$@")
  sleep 1 # XXX: race condition! hopefully, if it is not queued, it will start after one second.
  pueue --color=always status | grep '^\s*'"$task"' '  # XXX: hack: status query does not support filtering by ID
  pueue follow $task
}
```
but it's a little janky.
The comments show how difficult it is to mimic the behaviour, especially if we want user-friendly information about whether the new task is running or in a queue.

Myself, I just try to remember `pueue status` after every `pueue add`.

### Additional context

Thank you for the well-made project! :)

Nukesor:
Resolves #592


katrinafyi (approved):
Looks good! Thanks for working on this :)

Tested:
- follow works with fast command: `./pueue add -i --follow -- 'ls'`
- follow continues command in background on ctrl+c
- follow requires immediate flag to be set

I have a few very small comments. You can think about these if you want or ignore them.
- the error message when using `--follow` without `-i` might be unclear. it says "error: the following required arguments were not provided:  --immediate", but it's not obvious that --immediate is only required when --follow is given. i don't know if this is easy to change in the argument parsing library
- if you Ctrl+C while following, it might not be obvious that the task continues. for comparison, when pressing Ctrl+Z in a shell, it prints `^Zfish: Job 2, 'sleep 100' has stopped`
- i'm not sure if this was deliberate, but the new task ID is not shown. i think it's useful to know the ID for using with other commands. a message like `New task added (id 18).` (same as using `add` without follow)  could also help address the previous point.

Thanks again!

Nukesor:
1. That's a limit of the CLI library and I think the error is clear enough.
2. Just to be clear. Pueue **does not** aim to re-implement or replace the background task logic of the shell. I know that nushell uses it for that and that you're using it as a fancy `&` but that's not what it has been designed for.
   I'm already starting to get a bad feeling about this feature request, because this is exactly the type of user behavior that'll be attracted by this feature, leading to more feature requests of this type.
   I'll reconsider again whether I'll actually merge this.
3. The current behavior exactly matches the `--follow` behavior, which is to print the exact `stdout` output to the shell. Prepending any other output would potentially break pipes into other commands. E.g. somebody pipes the command's json output into `jq` or some follow-up command.

katrinafyi:
Re (3), I think the output should have the output of `add` then `follow`, since it does the tasks of both. I don't know if it makes sense to pipe `--follow` to another command, as the pipe would be broken if the command is Ctrl+C'ed. Printing the ID still lets you do a pipe by manually using `pueue follow ID | ...`.

Re (2), I understand.

Nukesor:
3. Hmm. I guess a good workaround to that would be to print all of Pueue's log output to `stderr`, which would be a good practice anyway.

Nukesor:
I changed the behavior to print the `add` output to `stderr` and the actual log output to `stdout` so the piping usecase is still viable.

I'll also merge this MR for the time being, but I don't plan to add any further features that aid Pueue mimicking as a shell background task manager. Stuff like that is the job of the shell itself :)

PR #604, opened 2025-02-10
Upgrade to `tracing` and `color_eyre` (2)

ActuallyHappening:
These changes make errors go from this:
image
To this:
image

And logs from this:
image
To this:
image

## Breaking changes:
Maybe `anyhow` was accidentally exposed somewhere? `color-eyre` was not exposed anywhere that `anyhow` wasn't, so for library usage this is a breaking change
The default logging implementation used only logs to stderr, I don't know if there was some logic in the old implementation to send some logs to stdout and some to stderr, but its now all consistent to stderr

## Checklist

Due to my unfortunately lacking `git` skills, there was an old PR with a branch with a rebase mistake I'm not bothered to learn how to fix [here](https://github.com/Nukesor/pueue/pull/602)

### DIFF
--- a/pueue/src/tracing.rs
+++ b/pueue/src/tracing.rs
@@ -1,0 +1,56 @@
+use crate::internal_prelude::*;
+use tracing::level_filters::LevelFilter;
+use tracing_subscriber::{
+    field::MakeExt, fmt::time::ChronoLocal, layer::SubscriberExt, util::SubscriberInitExt,
+    EnvFilter, Layer,
+};
+
+pub fn install_tracing(verbosity: u8) -> Result<()> {
+    let mut pretty = false;
+    let level = match verbosity {
+        0 => LevelFilter::WARN,
+        1 => LevelFilter::INFO,
+        2 => LevelFilter::DEBUG,
+        3 => LevelFilter::TRACE,
+        _ => {
+            pretty = true;
+            LevelFilter::TRACE
+        }
+    };
+
+    // tries to find local offset internally
+    let timer = ChronoLocal::new("%H:%M:%S".into());
+
+    type GenericLayer<S> = Box<dyn Layer<S> + Send + Sync>;
+    let fmt_layer: GenericLayer<_> = match pretty {
+        false => Box::new(
+            tracing_subscriber::fmt::layer()
+                .map_fmt_fields(|f| f.debug_alt())
+                .with_timer(timer)
+                .with_writer(std::io::stderr),
+        ),
+        true => Box::new(
+            tracing_subscriber::fmt::layer()
+                .pretty()
+                .with_timer(timer)
+                .with_target(true)
+                .with_thread_ids(false)
+                .with_thread_names(true)
+                .with_level(true)
+                .with_ansi(true)
+                .with_span_events(tracing_subscriber::fmt::format::FmtSpan::ACTIVE)
+                .with_writer(std::io::stderr),
+        ),
+    };
+    let filter_layer = EnvFilter::builder()
+        .with_default_directive(level.into())
+        .from_env()
+        .wrap_err("RUST_LOG env variable is invalid")?;
+
+    tracing_subscriber::Registry::default()
+        .with(fmt_layer.with_filter(filter_layer))
+        .with(tracing_error::ErrorLayer::default())
+        .init();
+
+    Ok(())
+}
--- a/pueue/tests/daemon/integration/edit.rs
+++ b/pueue/tests/daemon/integration/edit.rs
@@ -1,8 +1,8 @@
+use crate::prelude::*;
+
 use std::path::PathBuf;
 
-use anyhow::{bail, Result};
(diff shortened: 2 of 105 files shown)


pueue/src/tracing.rs
```diff
@@ -0,0 +1,56 @@
...
+        false => Box::new(
+            tracing_subscriber::fmt::layer()
+                .map_fmt_fields(|f| f.debug_alt())
+                .with_timer(timer)
+                .with_writer(std::io::stderr),
+        ),
+        true => Box::new(
+            tracing_subscriber::fmt::layer()
+                .pretty()
+                .with_timer(timer)
```
Nukesor, line 35:
```suggestion
                .map_fmt_fields(|f| f.debug_alt())
                .with_timer(timer)
```

pueue/tests/daemon/integration/edit.rs
```diff
@@ -32,7 +32,8 @@ async fn create_edited_task(shared: &Shared) -> Result<Vec<EditableTask>> {
 }
 
 /// Test if adding a normal task works as intended.
-#[test(tokio::test(flavor = "multi_thread", worker_threads = 2))]
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+#[tracing_test::traced_test]
```
Nukesor, line 36: I overlooked the wrapping `test` call in the previous review and didn't notice `test_log::test`, that's why this only pops up now.
`test_log` is a library that sets up logging in a testing environment so that one can easily use logging during testing. I currently no longer use it and apparently forgot to remove it.

Please remove `tracing_test` from the dependency tree as long as it isn't actively used. `test_log` can also just stay removed as it's no longer  actively used.

pueue/tests/daemon/integration/restart.rs
```diff
@@ -1,6 +1,7 @@
+use crate::prelude::*;
```
Nukesor, line 1: This is still called `prelude`

pueue_lib/src/process_helper/windows.rs
```diff
@@ -1,9 +1,9 @@
-// We allow anyhow in here, as this is a module that'll be strictly used internally.
+// We allow color_eyre in here, as this is a module that'll be strictly used internally.
 // As soon as it's obvious that this is code is intended to be exposed to library users, we have to
 // go ahead and replace any `anyhow` usage by proper error handling via our own Error type.
-use anyhow::{bail, Result};
+use color_eyre::{bail, Result};
 use command_group::GroupChild;
-use log::{error, info, warn};
+use tracing::{error, info, warn};
```
Nukesor, line 6: Can't you use the prelude in here?

Nukesor (changes requested):
Looking better, thanks :) 

A few nitpicks are left, but this should be soon good to go.

Could you go ahead and either fixup the latest commit into the first one or squash everything? I'm usually not too pedantic about properly scoped commits in my open source projects, but you're reverting my pueue_lib changelog commit in your first commit just to revert it later on again. Stuff like that just introduces unnecessary chatter in the git history which is something that should be avoided if possible.

Thanks :)

pueue/src/tracing.rs
ActuallyHappening: Unfortunately this method is not implemented for `pretty` (see here)[https://docs.rs/tracing-subscriber/latest/tracing_subscriber/field/trait.MakeExt.html#implementors], I'm pretty sure its implied from my own usage

pueue/tests/daemon/integration/edit.rs
ActuallyHappening: Yeah, I gotta admit, just searched for `tracing` equivalent, I just initialize my own logging but thought there was some special reason you weren't

pueue_lib/src/process_helper/windows.rs
ActuallyHappening: Yes we can, I didn't for clarity as the comment above mentioned `color_eyre::Result` I think, but good catch :)

pueue/tests/daemon/integration/restart.rs
ActuallyHappening: Yeah, done

ActuallyHappening:
OK, probably windows compiles.

Nukesor:
Welcome to multi-platform hell :no_mouth: 
Where you have to run either multiple OSs in parallel or wait for the CI pipeline every time :)

In case you want to iterate more quickly, reset your repos `main` onto my `main` and create a PR against your own repo. Then enable Actions for your fork and you should see the actions over there without having to wait for me to run them.

Nukesor:
Neat. Thanks for the contribution!

I fixed the last windows related issues for you and squashed the last two commits while at it :) 

Once the Pipeline succeeds, I'll go ahead and merge the MR!

ActuallyHappening:
Ah thanks! This was much more involved than I initially imagined ...

PR #630, opened 2025-03-17
Fix extraneous double quotes in Windows service

MolotovCherry:
When running `pueued -c "space path" service install`, the service cli got expanded to `pueued --config "\"space path\"" service run`, which is definitely not correct. Turns out that the `windows_service` crate already handled the case of spaces in the arg for us. Nice. This PR fixes this bug. Now it expands to `pueued --config "space path" service run` or `pueued --config no_space_path service run`

(I don't know the next version for the changelog, but I assumed it's 4.0.1. Please let me know if the changelog entry needs to be fixed)

TODO:
- [ ] double check if this area is correct: (I think it is, but gotta make sure) https://github.com/Nukesor/pueue/blob/33290e674212593e6306ca481e868be6f0a19404/pueue/src/daemon/service.rs#L634

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## \[4.0.1\] - unreleased
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


Nukesor:
4.0.1 is correct :+1:  

I'll release a patch version for it soonish then :)

MolotovCherry:
> Thanks, 4.0.1 is also correct :+1:  
> 
> I'll release a patch version for it soonish then :)

Could we hold off on the next patch release for a moment? I have one more related area I want to be certain is correct

PR #631, opened 2025-03-19
Fix flags causing service to not start.

MolotovCherry:
This is a follow up to #630 . After checking, flags indeed were causing the service start up fail. I have now tested them properly and fixed the issue.

The issue was that the rust binary expects the first arg to be the current exe, e.g. `"C:\my-current-exe.exe"`, but I was not giving it.

This full expanded command now works (including profile too if specified).
`R:\path\to\pueued.exe --config C:\path\to\pueue.yml service run`

I did a small refactor in the process, cause why not

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Fixed
 
 - Fix extraneous double quotes being added to --config and --profile flags in Windows service install. #630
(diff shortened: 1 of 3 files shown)


MolotovCherry:
Still a draft at the moment. I'll mark it ready once I feel I'm good to proceed.

MolotovCherry:
I have done some testing, and it appears to work properly. Should be ready to go! 😄

Nukesor:
Awesome, thanks :) 

In it goes.

PR #650, opened 2025-12-12
feat(status): display duration

FrancescElies:
**Problem** when looking at `pueue status` it takes too many brain cycles to figure out how long a task took.

**This pr** adds duration column to `pueue status`
image

### Checklist

### DIFF
--- a/pueue/src/format.rs
+++ b/pueue/src/format.rs
@@ -1,13 +1,88 @@
-use chrono::{DateTime, Local};
+use chrono::{DateTime, Local, TimeDelta};
 
 use pueue_lib::settings::Settings;
 
 // If the enqueue at time is today, only show the time. Otherwise, include the date.
 pub fn format_datetime(settings: &Settings, enqueue_at: &DateTime<Local>) -> String {
     let format_string = if enqueue_at.date_naive() == Local::now().date_naive() {
-        &settings.client.status_time_format
+        &settings.client.get_status_time_format()
     } else {
-        &settings.client.status_datetime_format
+        &settings.client.get_status_datetime_format()
     };
     enqueue_at.format(format_string).to_string()
+}
+
+pub fn humanize_duration(duration: TimeDelta) -> String {
+    let days = duration.num_days();
+    let leftover = duration - TimeDelta::days(days);
+
+    let hours = leftover.num_hours();
+    let leftover = leftover - TimeDelta::hours(hours);
+
+    let minutes = leftover.num_minutes();
+    let leftover = leftover - TimeDelta::minutes(minutes);
+
+    let seconds = leftover.num_seconds();
+    let leftover = leftover - TimeDelta::seconds(seconds);
+
+    let millis = leftover.num_milliseconds();
+
+    let mut parts = Vec::new();
+    if days > 0 {
+        parts.push(format!("{days}day"));
+    }
+    if hours > 0 {
+        parts.push(format!("{hours}hr"));
+    }
+    if minutes > 0 {
+        parts.push(format!("{minutes}min"));
+    }
+    if seconds > 0 {
+        parts.push(format!("{seconds}sec"));
+    }
+    if days == 0 && hours == 0 && minutes == 0 && seconds < 3 && millis > 0 {
+        parts.push(format!("{millis}ms"));
+    }
+
+    parts.join(" ")
+}
+
+#[cfg(test)]
+mod test {
+    use super::humanize_duration;
+    use chrono::TimeDelta;
+
+    macro_rules! assert_eq_seconds {
+        ($seconds:expr, $str: expr) => {
+            assert_eq!(
+                humanize_duration(TimeDelta::new($seconds, 0).unwrap()),
+                $str.to_owned()
+            );
+        };
+    }
+
+    macro_rules! assert_eq_millis {
+        ($millis:expr, $str: expr) => {
+            assert_eq!(
+                humanize_duration(TimeDelta::try_milliseconds($millis).unwrap()),
+                $str.to_owned()
+            );
+        };
+    }
+
+    #[test]
+    fn duration() {
+        assert_eq_millis!(500, "500ms");
+        assert_eq_millis!(1500, "1sec 500ms");
+        assert_eq_millis!(2500, "2sec 500ms");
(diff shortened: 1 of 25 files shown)


pueue/src/format.rs
```diff
@@ -11,3 +11,38 @@ pub fn format_datetime(settings: &Settings, enqueue_at: &DateTime<Local>) -> Str
     };
     enqueue_at.format(format_string).to_string()
 }
+
+pub fn humanize_duration(d: TimeDelta) -> String {
```
FrancescElies: Since I don't know if this pr is eve desired, I intentionally avoided overhead and didn't add any tests for this function.

Considered using [chrono-humanize](https://docs.rs/chrono-humanize/latest/chrono_humanize/) but didn't want to add a dependency without asking.

FrancescElies:
I agree, one more column here now, another one tomorrow because of something else and soon the table is too bloated.

Personally I feel we could swap The `End` column with `Duration`, I normally care when a task started and how long it took, real end time is secondary for me, but I understand if you don't feel that way.

If you are ok with the swap we would stay with the same number of columns and do the other suggestions in separate prs.
 
With regards to your `status` subsection i can go and do that and see how it feels either in this pr or a separate one.

Another option is to add duration only to the `pueue status --json` (which now I realize I didn't) and let the user select what it wants, using `jq` or similar and then making your own alias , e.g.
image

Nukesor:
Heyo, I'll probaly only get to reviewing this in the new year :D
The last month was pretty hectic and I'm away for the next few days :)

pueue/src/format.rs
Nukesor: I'm ok with `chrono-humanize` :).
Although, I'm wondering, whether we should have a set format.

Personally, I would like to have the duration be as compact as possible. Something along the line of `1 days 05:35:45`.

From what I've seen, there's no strftime-like API for Duration, which is a bit annoying. In practice, it shouldn't be too hard to build this ourselves (with a limited set of variables), since we already have `handlebars` templating in pueue due to the callback logic. With that, we could provide a simple strftime-like `duration_format` configuration option on top. Maybe even a `days_duration_format` for very long tasks?

pueue_lib/tests/settings_backward_compatibility.rs
```diff
@@ -23,8 +23,42 @@ mod tests {
...
+        assert_eq!(
+            settings.client.status.datetime_format,
+            "mock402-%Y-%m-%d\n%H:%M:%S"
+        );
+        assert_eq!(
+            settings.client.status.additional_columns,
+            vec!["Duration".to_owned(),]
+        );
 
         assert!(config_found);
```
Nukesor: Since we're not introducing a new major version, this test isn't necessary. We just need to ensure backwards compatibility to the oldest supported version.

Just remove it :)

pueue_lib/src/settings.rs
```diff
@@ -112,21 +113,58 @@ pub struct Client {
...
+    pub max_lines: Option<usize>,
+    /// The format that will be used to display time formats in `pueue status`.
+    #[serde(default = "default_status_time_format")]
+    pub time_format: String,
+    /// The format that will be used to display datetime formats in `pueue status`.
+    #[serde(default = "default_status_datetime_format")]
+    pub datetime_format: String,
+    #[serde(default = "default_additional_columns")]
+    pub additional_columns: Vec<String>,
+}
```
Nukesor: I have a proposal to ease the transition.

How about, we make all of these optional?
We then expose getters for these fields on `Settings`, which return the `Status` configs (if set), with a fallback to the old configs.

When the old config options are removed in the future, we then simply return the `default_status_datetime_format`, etc. when a config value is `None`.

-> We move the default value logic from the deserialization step to a function.

This should make the whole replicate_deprecated_fields logic unnecessary and make this thing easier.

Nukesor (changes requested):
Neat, thanks for your work :) 

I have a few improvements in mind, but the overall thing already looks pretty good!

FrancescElies:
just rebased and solved merge conflict, nothing new yet.

pueue_lib/tests/data/v4.0.2_settings.yml
```diff
@@ -0,0 +1,29 @@
...
+client:
+  bogus_settings: ~
+  restart_in_place: false
+  read_local_logs: true
+  show_confirmation_questions: false
+  status:
+    show_expanded_aliases: true
+    max_lines: 42
+    time_format: "mock402-%H:%M:%S"
+    datetime_format: "mock402-%Y-%m-%d\n%H:%M:%S"
```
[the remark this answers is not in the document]
FrancescElies: because of v4.0.2, probably a bit dumb.

Given that we are removing the previous test we don't need that file either, thus removing it too.

pueue_lib/src/settings.rs
FrancescElies: sounds good I will give it a go

FrancescElies:
While not very rigorously I shortly tested with the config below. 
I am not using pueue daily that much anymore :/

I like pueue, the main reason why I discovered it is because sometimes I have to use windows and would like to use same tools across different OS, on windows  nowadays there is also a terminal multiplexer (afaict  was not the case before). 
This is why  lately i have sadly seen myself using pueue less and less.


``` yaml
# pueued.yml
# ---
client:
  status:
    additional_columns: ["Duration"]
``` 

2026-06-13T09:31:15

lovehumans:
But is your PR basically finished and could be accepted?

FrancescElies:
If the maintainers of this project are fine with the last commits I pushed after the first review. Then yes.

Unless I missed something,I believe I address all open points

Nukesor:
I justed a new work project and am gone for a few days on-and-off in the next week. I'm not sure when I'll have the time to review the MR.

Also, in case you haven't seen, the project is somewhat in a [feature freeze](https://github.com/Nukesor/pueue/issues/669).
I'll consider this MR anyway though.

FrancescElies:
I will be gone now for two weeks.  Even if this pr doesn’t get merged that’s fine with me, i just did it to learn a bit about the internals of pueue. 
For the duration addition i can always build this branch from source :)

PR #666, opened 2026-03-06
feat: add --all and --group options to restart, remove, and edit commands

### LINKED ISSUE #656: restart all tasks with `pueue restart all`

i have many task to be restart, i need restart them type 

pueue restart 0 1 ... 100 

i want a way to restart all tasks without provide task ids

restart all tasks many times 

### Alternatives

_No response_

### Additional context

_No response_

gitduk:
## Summary
  Add flexible filtering options to batch operations for `restart`, `remove`,
  and `edit` commands.

  ## Changes
  - **restart**: Add composable `--all`, `--group`, `--failed` options
  - **remove**: Add `--all` and `--group` options
  - **edit**: Add `--all` and `--group` options

  The `--all` and `--group` options are mutually exclusive. The `--failed`
  option can be combined with either `--all` or `--group` to filter only failed
   tasks.
   
   Fixes #656 

  ## Examples
  ```bash
  # Restart all finished tasks
  pueue restart --all

  # Restart all finished tasks in work group
  pueue restart --group work

  # Restart all failed tasks
  pueue restart --all --failed

  # Remove all finished tasks in work group
  pueue remove --group work

  # Edit all stashed/queued tasks
  pueue edit --all

  Testing

  - All 136 tests pass
  - Clippy clean
  - Code formatted

## Checklist


pueue/src/client/commands/remove.rs
```diff
@@ -9,7 +9,32 @@ pub async fn remove(
...
+        } else {
+            state.filter_tasks(done_filter, None)
+        };
+
+        if filtered_tasks.matching_ids.is_empty() {
+            println!("No finished tasks to remove.");
+            return Ok(());
+        }
+
+        filtered_tasks.matching_ids
```
Nukesor: This got me thinking.

We have three commands to remove tasks right now.

`pueue reset [--groups]` removes **everything**, including killing running tasks (optionally filtered by group).
`pueue clear [--group] [--successful-only]` Removes all finished tasks (or of a group), optionally only removing successful tasks.

The new logic of `remove` basically behaves like `pueue clear`, but with less configuration options.
To make it worth its own behavior, `remove --all` should really remove **all** tasks that aren't running rigght now, including stashed, queued.

The question is, if that makes any sense at all. Right now, I don't really have a usecase for that in mind, except maybe "I just want to clear my queue, but let the running processes finish?". An alternative way to achieve this would be to `pueue pause --wait && pueue wait && pueue reset`

What do you think? Should we add that behavior or just revert the addition of the `--all` flag on `pueue remove`?

pueue/src/client/commands/restart.rs
```diff
@@ -48,46 +49,49 @@ pub async fn restart(
...
+                            result: TaskResult::Success,
+                            ..
+                        }
+                    )
+                })
+                .collect()
+        } else {
+            // No --failed: all finished tasks
+            filtered_tasks.matching_ids
+        };
```
Nukesor: You can condense this block into the `done_filter` (probably rename it to while at it :D).
There is an `task.failed()` function to check whether a task is failed or not.
That way you don't have to iterate through the tasks twice.

Nukesor (changes requested):
Neat, thanks for the contribution!

I really like the way the `restart` flags look right now. Composability is way better than those previous "pre-composed" flags.

I have a few comments, one general question we should think about.

Beside that, I would really like to see some tests! The fact that nothing breaks, even though we introduce braking changes, shows that this needs to be urgently tested.
For these changes, I would be ok with having some "simplistic" tests, where there's a single task that matches a condition, and another that doesn't :) 

Check out `pueue/tests/client/integration/restart` on how such client tests would look like. The gist is: We have a clean running `daemon` in the background per test. The required state for the test in question is created via direct API calls (e.g. `add_task`, `wait_for_task_condition`, etc., which are effectively test helpers/fixtures).
Once that initial state is reached, the actual `pueue` binary is called. Afterwards, the state is checked, whether it was modified in the expected way. 

Thanks again, looking forward to seing this merged!

Nukesor (changes requested):
One other thing.
The texts/comments are a bit off at times at and the failed task filter logic is a bit off as well.

In case you're using an LLM, please disclose this in your PR description. It's necessary so the expectations are clear and I can perform stricter reviews. All LLM generated MRs I've yet received across all my projects contained weird comments, wrong assumptions or small verbose algorithmic issues, just like that filter logic, so it sadly is a real necessity.

In case this wasn't LLM assisted, sorry, but I had to ask.

Nukesor (changes requested):
Obvious LLM generated comments/answers despite the guidelines stated in the AGENTS.md.

I need to chill for a bit and think about, whether I should even consider to merge this MR.

pueue/src/client/commands/remove.rs
gitduk: This isn't an LLM answer. The above solution is something I haven't fully understood. 

I'm thinking that three subcommands for clearing tasks are too many. Since a flag can be used to mark the task status, a single `remove` subcommand is sufficient for all scenarios. `reset` can be replaced with `remove --all`. To clear tasks in the completed state, you can use `remove --finished` or `remove --failed --successed` or some other flags, which is simple and clear. To remove a task in a certain state, use `remove + flag`; to remove a specific task, use `remove + id`.

I want to revert the `--all` commit, remove the `reset` and `clear` subcommands, and redesign the task filtering mode.

gitduk:
> Obvious LLM generated comments/answers despite the guidelines stated in the AGENTS.md. I need to chill for a bit and think about, whether I should even consider to merge this MR.

I apologize, but this code was indeed generated by LLM. I think not merging it is a good idea, but I hope you can consider simplifying the command system in version 5.0, as it is a bit complex.

gitduk:
> > Obvious LLM generated comments/answers despite the guidelines stated in the AGENTS.md. I need to chill for a bit and think about, whether I should even consider to merge this MR.
> 
> I apologize, but this code was indeed generated by LLM. I think not merging it is a good idea, but I hope you can consider simplifying the command system in version 5.0, as it is a bit complex.

I don't have time to read the pueue source code right now, so I'll close this PR for now. I need to figure out how to simplify the subcommand system and code without llm before resubmitting.

